PR #2, opened 2015-04-15
Fix bug while calculating the periods of a task related to the duration ...

astahlhofen:
The calculation of the periods executed by a task related to the duration of the test is wrong. If you want to convert seconds to microseconds, then you have to multiply with 1e6 (=1.000.000) instead of 10e6 (=10.000.000). Only a small fix ;-)
Best regards,
Andreas

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+*~
+.*.swp
 COPYING
 Makefile
 Makefile.in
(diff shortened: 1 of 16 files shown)


vingu-linaro:
This pull request doesn't seem to be relevant anymore. Maybe we can close it ?

astahlhofen:
Yes, I can confirm your suggestion :-)

PR #5, opened 2015-12-04
Add json files for EASv5 profiling on Hikey

Leo-Yan:
Signed-off-by: Leo Yan <leo.yan@linaro.org>

### DIFF
--- a/doc/examples/eas_v5_hikey/rt-app-13.json
+++ b/doc/examples/eas_v5_hikey/rt-app-13.json
@@ -1,0 +1,24 @@
+{
+    "tasks": {
+        "thread0": {
+            "instance": 8, 
+            "loop": -1, 
+            "run": 260, 
+            "sleep": 0, 
+            "timer": {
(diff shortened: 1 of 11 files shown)


jlelli:
Hi,
test cases are automatically generate using LISA.
I think is better if we don't add this task set to rt-app and point people to use LISA instead.
Is that OK for you?
Thanks!

Leo-Yan:
Hi Juri,

On Sat, Feb 13, 2016 at 08:22:18AM -0800, Juri Lelli wrote:
> Hi,
> test cases are automatically generate using LISA.
> I think is better if we don't add this task set to rt-app and point people to use LISA instead.
> Is that OK for you?

Yes, absolutely. I also will directly work on LISA and drop my own
rt-app scripts.

Thanks,
Leo Yan

jlelli:
Hi!
Great! Thanks a lot for your help and support!
Best,
Juri

PR #15, opened 2016-12-12
Add Barrier Sync Primitive

credp:
I wanted a way to provide a sync point for two threads where it didn't
matter which one got there first - the first thread to reach the sync
point would wait, and the last would just signal and continue.

Using a bare lock/signal/wait/unlock on both sides results in the second
thread being suspended briefly immediately after it signals which is
not desirable when trying to mimic existing thread activation patterns.

If we don't use bare signal/wait then it doesn't need to be restricted
to just 2 threads, you can sync an arbitrary number (>1) of threads at
a given point in a use case by including barrier events in them.

This effect could be achieved with a pthread_barrier, but there is no
officially sanctioned way to cancel a barrier sync event once it has
started and no way to tell if any threads are waiting without adding
such a mechanism externally, so we use the pattern here instead.

Each barrier point has a mutex, a shared counter and a condition
variable.

When parsing the usecase, the shared counter is incremented for
each reference.

At barrier sync, the event locks the event-specific mutex and checks
to see if it is the last task. If it is not the last task, then it
waits on the event-specific condition variable. If it is the last
task, then it sends a broadcast to all other users of the condition
variable. The shared counter is used to determine when the last
task reaches the sync point.

Signed-off-by: Chris Redpath <chris.redpath@arm.com>

### DIFF
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -422,6 +422,39 @@ following sequence:
 	"unlock" : "mutexA"
 }
 
+* barrier : String. Used as at least a pair where the name must match.
+Any number of matching uses will cause all threads hitting the barrier event
+to wait for a signal. The number of users is recorded, so that when the last user
+hits the barrier event, that thread will broadcast and continue to the next
+step. This is conceptually exactly the same as a pthread_barrier_wait operation
+however, using a pthread_barrier would impose some strict conditions on usage
+around thread cleanups - primarily that you cannot cancel an in-progress barrier
+operation which would mean that we have to restrict cleanup to only be possible
+at the end of a loop cycle (i.e. all phases are complete). This would be too
+restrictive for most uses so here we use an alternative.
+
+In this implementation, the barrier event manages its own mutex and uses a
+variable shared between users to track waiting tasks, protected by the mutex.
+
+You must use a unique name for each sync event since the number of users is
+taken from the number of references to the name in the input json. i.e. each
+name must represent a single sync point, and be shared amongst all threads
+which wish to syncronise at that point.
+
+The barrier event "barrier" : "SyncPointA"
+generates the following sequence:
+{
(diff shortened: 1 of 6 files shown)


doc/tutorial.txt
```diff
@@ -422,6 +422,39 @@ following sequence:
 	"unlock" : "mutexA"
 }
 
+* barrier : String. Used as at least a pair where the name must match.
```
jlelli: Do we need to check that we have at least two users of a barrier while parsing the json file?
If not the only user can get stuck, can't it?

jlelli:
Chris, could you please also post this to sched-tools mailing list?
It would be good that more eyes can see it before we merge it, IMHO.
Apart from the comment I made (and a trivial merge conflict resolution), doesn't look to far from being OK!

doc/tutorial.txt
credp: Yes, if you only have one user that user is guaranteed to get stuck there. I will add that.

credp:
No problem - I'll fix the conflicts up, add a check for >1 users of a primitive, and post it up to sched-tools mailing list.

doc/tutorial.txt
credp: Actually, the way I implemented it I think if there's only one reference you'll just emit a signal and carry on. I will test that it actually happens that way...

credp: Yep, it works just fine if you have only one user of a barrier sync.

jlelli: Great! Maybe add some comments about this in the documentation so that the behaviour is clear?

credp:
Vincent gave me a verbal ACK at Connect BUD17, so I'm merging this.

vingu-linaro:
Yes. I confirm the Ack.
Sorry i forgot to log in github and ack it

PR #16, opened 2016-12-26
rt-app: add support for "yield" event

ShayDamir:
"yield" event simply calls pthread_yield()
This has a special meaning for SCHED_DEADLINE tasks


vingu-linaro:
Merged.
Thanks

PR #17, opened 2017-01-25
Remove unsued cpuset in parse_thread_data()

deggeman:
No need to retrieve the cpuset via json_object_get_array(). We use
json_object_array_get_idx() instead to get the cpus of the cpuset
instead.

Signed-off-by: Dietmar Eggemann <dietmar.eggemann@arm.com>
Acked-by: Juri Lelli <juri.lelli@arm.com>


vingu-linaro:
looks good to me
Acked-by: Vincent Guittot <vincent.guittot@linaro.org>

PR #18, opened 2017-01-26
doc: Fix ./configure command in tutorial.txt

bjackman:
With the provided command I get this error from ./configure:

```
checking whether the C compiler works... no
configure: error: in `/home/brendan/sources/rt-app':
configure: error: C compiler cannot create executables
See `config.log' for more details
```

This seems to be because the "libs/" part is bogus.


vingu-linaro:
yes it should have been -L.libs/ but the latter seems to be empty so we can remove it

PR #21, opened 2017-03-08
Enhancing debug prints

credp:
I've been trying to implement more complex use cases using rt-app primitives, and the existing debug infrastructure doesn't tell you that much about what's happening at playback time.

These patches let you see the names of shared resources and the thread indices so that you can figure out why your use case json doesn't run to completion.

I also added the ftrace output (when debug is enabled and ftrace channel is open) to make debugging using ftrace easier.

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -406,13 +406,15 @@ int run(int ind,
 	int nbevents = pdata->nbevents;
 	int i, lock = 0;
 	unsigned long perf = 0;
+	char type_name[64];
 
 	for (i = 0; i < nbevents; i++)
 	{
 		if (!continue_running && !lock)
 			return perf;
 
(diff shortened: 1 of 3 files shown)


jlelli:
Hey, didn't notice this until now! It's really useful, thanks.
Could you please rebase (should be only minor conflict) and consider these changes?
[0001-Suggested-changes-for-Chris.txt](https://github.com/scheduler-tools/rt-app/files/1097651/0001-Suggested-changes-for-Chris.txt)

credp:
OK, I'll take a look.

jlelli:
As discussed, I've taken your patches and created a new [PR](https://github.com/scheduler-tools/rt-app/pull/32).
Moving the discussion there and closing this.

PR #27, opened 2017-06-08
Fix loop property handling

jlelli:
loop property handling appears to be broken (i.e., loop at thread level is ignored and always defaults to -1 when there isn't a phases object).
Fix it by also sanitising code a bit.

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -600,7 +600,7 @@ void *thread_body(void *arg)
 	pid_t tid;
 	struct sched_attr attr;
 	unsigned int flags = 0;
-	int ret, i, j, loop, idx;
+	int ret, phase, phase_loop, thread_loop, log_idx;
 
 	/* Set thread name */
 	ret = pthread_setname_np(pthread_self(), data->name);
@@ -753,16 +753,28 @@ void *thread_body(void *arg)
 				NULL);
 	}
 
-	i = j = loop = idx = 0;
+	/*
+	 * phase        - index of current phase in data->phases array
+	 * phase_loop   - current iteration of current phase
+	 *                (corresponds to "loop" at phase level)
+	 * thread_loop  - current iteration of thread/phases
+	 *                (corresponds to "loop" at task level
+	 * log_idx      - index of current row in the log buffer
+	 */
+	phase = phase_loop = thread_loop = log_idx = 0;
 
-	while (continue_running && (i != data->loop)) {
+	/* The following is executed for each phase. */
+	while (continue_running && thread_loop != data->loop) {
 		struct timespec t_diff, t_rel_start;
 
 		set_thread_affinity(data, &pdata->cpu_data);
 
 		if (opts.ftrace)
-			log_ftrace(ft_data.marker_fd, "[%d] begins loop %d phase %d step %d", data->ind, i, j, loop);
-		log_debug("[%d] begins loop %d phase %d step %d", data->ind, i, j, loop);;
+			log_ftrace(ft_data.marker_fd,
+				   "[%d] begins thread_loop %d phase %d phase_loop %d",
+				   data->ind, thread_loop, phase, phase_loop);
+		log_debug("[%d] begins thread_loop %d phase %d phase_loop %d",
+			  data->ind, thread_loop, phase, phase_loop);
 
 		memset(&ldata, 0, sizeof(ldata));
 		clock_gettime(CLOCK_MONOTONIC, &t_start);
@@ -771,7 +783,7 @@ void *thread_body(void *arg)
 		clock_gettime(CLOCK_MONOTONIC, &t_end);
 
 		if (timings)
-			curr_timing = &timings[idx];
+			curr_timing = &timings[log_idx];
 		else
 			curr_timing = &tmp_timing;
 
@@ -794,26 +806,37 @@ void *thread_body(void *arg)
 			log_timing(data->log_handler, curr_timing);
 
 		if (opts.ftrace)
-			log_ftrace(ft_data.marker_fd, "[%d] end loop %d phase %d step %d",
-				   data->ind, i, j, loop);
+			log_ftrace(ft_data.marker_fd,
+				   "[%d] end thread_loop %d phase %d phase_loop %d",
+				   data->ind, thread_loop, phase, phase_loop);
 
-		loop++;
-		if (loop == pdata->loop) {
-			loop = 0;
+		phase_loop++;
+		/* Reached the specified number of loops for this phase. */
+		if (phase_loop == pdata->loop) {
+			phase_loop = 0;
 
-			j++;
-			if (j == data->nphases) {
-				j = 0;
-				i++;
+			phase++;
+			if (phase == data->nphases) {
+				/*
+				 * Phases are finished, but we might potentially have
+				 * to start over (depending on data->loop).
+				 */
+				phase = 0;
+				thread_loop++;
+				/*
+				 * Overflow due to data->loop being -1 (looping forever)
+				 * Reset thread_loop, so that we can continue looping.
+				 */
+				if (thread_loop < 0)
+					thread_loop = 0;
 			}
-
-			pdata = &data->phases[j];
+			pdata = &data->phases[phase];
 		}
 
-		idx++;
-		if (idx >= timings_size) {
+		log_idx++;
+		if (log_idx >= timings_size) {
 			timing_loop = 1;
-			idx = 0;
+			log_idx = 0;
 		}
 	}
 
@@ -828,9 +851,11 @@ void *thread_body(void *arg)
 	}
 
 	if (timings) {
-		for (j = idx; timing_loop && (j < timings_size); j++)
+		int j;
+
+		for (j = log_idx; timing_loop && (j < timings_size); j++)
 			log_timing(data->log_handler, &timings[j]);
-		for (j = 0; j < idx; j++)
+		for (j = 0; j < log_idx; j++)
 			log_timing(data->log_handler, &timings[j]);
 	}
 
--- a/src/rt-app_parse_config.c
+++ b/src/rt-app_parse_config.c
@@ -799,6 +799,7 @@ parse_thread_data(char *name, struct json_object *obj, int index,
 		data->phases = malloc(sizeof(phase_data_t) * data->nphases);
 		foreach(phases_obj, entry, key, val, idx) {
 			log_info(PIN "Parsing phase %s", key);
+			assure_type_is(val, phases_obj, key, json_type_object);
 			parse_thread_phase_data(val, &data->phases[idx], opts, (long)data);
 		}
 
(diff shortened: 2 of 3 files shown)


punitagrawal:
I took the fixes/loops branch for a spin and it behaved as expected, i.e., the workload stopped after executing the specified number of loops.

src/rt-app_parse_config.c
```diff
@@ -811,7 +811,7 @@ parse_thread_data(char *name, struct json_object *obj, int index,
 		data->phases = malloc(sizeof(phase_data_t) * data->nphases);
 		parse_thread_phase_data(obj,  &data->phases[0], opts, (long)data);
 		/* Get loop number */
-		data->loop = -1;
+		data->loop = get_int_value_from(obj, "loop", TRUE, -1);
 	}
```
vingu-linaro: This is called when there is no "phase" object defined in the "thread" object. In this case, we assume that events and loop of the phase are put directly in the "thread".
In this case we have only 1 "loop" key value for phase and thread and as we call parse_thread_phase_data with thread object, the "loop" will already be taken in the phase.
With this change, the "loop" key will be used twice. Once in the parse_thread_phase_data and once in parse_thread_data.

so if we have
"thread0" : {
			"instance" : 1,
			"loop" : 2,
			"run" :   10000,
}

we will run 4 time the run events

vingu-linaro:
sorry i made a mistake i wanted to approved the 1 commit not the whole series that i haven't finished to review

src/rt-app_parse_config.c
jlelli: Mmm, I expected the same, but I'm pretty sure it wasn't working as expected, i.e. was looping indefinitely. I also modified the while(continue_running) loop in a subsequent patch to only consider thread object loop when there is only a single phase (or none).

jlelli:
No problem.

src/rt-app_parse_config.c
vingu-linaro: Originally it was: data->loop = 1  instead of -1
and everything was working fine when phase was merged in the thread
But viresh raises a problem which i had reproduced but can't reproduce now

jlelli: With Viresh' fix only (checking out previous commit of my series( and the following json file the behaviour seems to be loop indefinitely ?
```
{
        "tasks" : {
                "thread0" : {
                        "instance" : 1,
                        "cpus"     : [0],
                        "run"      : 1000000,
                        "sleep"    : 100000,
                        "loop"     : 3
                }
        },
        "global" : {
                "duration" : -1,
                "calibration" : 585,
                "default_policy" : "SCHED_OTHER",
                "pi_enabled" : false,
                "lock_pages" : false,
                "logdir" : "./",
                "log_size" : 10,
                "log_basename" : "rt-app"
        }
}
```

vingu-linaro: I have asked viresh about the config that was behaving wrongly without his fix to get the full picture but we might need to revert his fix

jlelli: Right. So, if I revert Viresh' fix and use the following json, the thread only executes once and then terminates (while it should execute indefinetely)
```
{
        "tasks" : {
                "thread0" : {
                        "instance" : 1,
                        "cpus"     : [0],
                        "run"      : 1000000,
                        "sleep"    : 100000
                }
        },
        "global" : {
                "duration" : -1,
                "calibration" : 585,
                "default_policy" : "SCHED_OTHER",
                "pi_enabled" : false,
                "lock_pages" : false,
                "logdir" : "./",
                "log_size" : 10,
                "log_basename" : "rt-app"
        }
}
```
I guess that was the problem.
The final result of this series should fix both cases.

vingu-linaro: i think that commit "parse_config: correctly parse thread object loop number" is useless.
Then i have some concerns to discard the thread loop when there is only 1 phase. This create a special case in which you can easily fall in when you update/change your json file.

vingu-linaro: the problem is : 
with json like below
"thread0" : {
"run" : 10000,
}
phase->loop will be 1 and thread->loop must be -1 to loop indefinitely

with json like below
"thread0" : {
"loop": X
"run" : 10000,
}
phase->loop will be X and thread->loop must be 1 to loop only X times.

AFAICT, there is no problem as soon as we have a phase object, isn't it ?

So may be do:
@@ -811,7 +811,11 @@ parse_thread_data(char *name, struct json_object *obj, int index,
 		data->phases = malloc(sizeof(phase_data_t) * data->nphases);
 		parse_thread_phase_data(obj,  &data->phases[0], opts, (long)data);
 		/* Get loop number */
-		data->loop = get_int_value_from(obj, "loop", TRUE, -1);
+		if(get_in_object(obj, "loop", TRUE))
+			data->loop = 1;
+		else
+			data->loop = -1;
+
 	}
 
 }

vingu-linaro: so we don't have to discard thread->loop if there is only one phase

jlelli: But with this set we actually discard phase->loop and not thread->loop when there is only a single phase.
So with the below
```
{
        "tasks" : {
                "thread0" : {
                        "instance" : 1,
                        "cpus" : [0],
                        "loop" : 3,
                        "phases" : {
                                "phase1" : {
                                        "loop" : 4,
                                        "run" :  1000000
                                }
                        }
                }
        },
        "global" : {
                "duration" : -1,
                "calibration" : 585,
                "default_policy" : "SCHED_OTHER",
                "pi_enabled" : false,
                "lock_pages" : false,
                "logdir" : "./",
                "log_basename" : "rt-app"
        }
}
```
We will be looping for 3 times, instead of 12, which is a sane thing to do?

jlelli: Maybe we could add a warning for this special case?

vingu-linaro: No, my proposal only applies to the else part of if (phases_obj) when there is no phase object

jlelli:
On 08/06/17 08:29, vingu-linaro wrote:
> vingu-linaro commented on this pull request.
> 
> 
> 
> > @@ -811,7 +811,7 @@ parse_thread_data(char *name, struct json_object *obj, int index,
>  		data->phases = malloc(sizeof(phase_data_t) * data->nphases);
>  		parse_thread_phase_data(obj,  &data->phases[0], opts, (long)data);
>  		/* Get loop number */
> -		data->loop = -1;
> +		data->loop = get_int_value_from(obj, "loop", TRUE, -1);
>  	}
> 
> the problem is : 
> with json like below
> "thread0" : {
> "run" : 10000,
> }
> phase->loop will be 1 and thread->loop must be -1 to loop indefinitely
> 
> with json like below
> "thread0" : {
> "loop": X
> "run" : 10000,
> }
> phase->loop will be X and thread->loop must be 1 to loop only X times.
> 
> AFAICT, there is no problem as soon as we have a phase object, isn't it ?
> 
> So may be do:
> @@ -811,7 +811,11 @@ parse_thread_data(char *name, struct json_object *obj, int index,
>  		data->phases = malloc(sizeof(phase_data_t) * data->nphases);
>  		parse_thread_phase_data(obj,  &data->phases[0], opts, (long)data);
>  		/* Get loop number */
> -		data->loop = get_int_value_from(obj, "loop", TRUE, -1);
> +		if(get_in_object(obj, "loop", TRUE))
> +			data->loop = 1;
> +		else
> +			data->loop = -1;
> +
>  	}
>  
>  }
> 

This works for tasks that have no phases, but the following seems to
loop forever?

{
	"tasks" : {
		"thread0" : {
			"instance" : 1,
			"cpus" : [0],
			"phases" : {
				"phase1" : {
					"loop" : 3,
					"run" :  1000000
				}
			}
		}
	},
	"global" : {
		"duration" : -1,
		"calibration" : 585,
		"default_policy" : "SCHED_OTHER",
		"pi_enabled" : false,
		"lock_pages" : false,
		"logdir" : "./",
		"log_basename" : "rt-app"
	}
}

jlelli:
@vingu-linaro I tried to implement your idea. Please have a look.

vingu-linaro:
For your example, it's fine to loop for ever as the default value of loop at thread level is -1 (infinite)

jlelli:
Oh, right. So, my refactoring/renaming of variables "sanitize and fix loop" shuold keep this behaviour, but IMHO cleans up code a bit. Does it look OK to you?

src/rt-app.c
```diff
@@ -794,26 +803,44 @@ void *thread_body(void *arg)
...
+				 * - perform a number of task level iterations of
+				 *   the same phase(s)
+				 *
+				 * In both cases, start over from first phase.
+				 *
+				 */
+				phase_loop = 0;
+				phase = 0;
+				thread_loop++;
+				pdata = &data->phases[0];
```
vingu-linaro: you can move "pdata = &data->phases[phase];"  after if and else if condition and remove "pdata = &data->phases[0];" from the else if

jlelli: I think we can only do something like

```
index a7f26c9..1a46c83 100644
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -817,7 +821,6 @@ void *thread_body(void *arg)
 				 * (and reset loop count for it)
 				 */
 				phase_loop = 0;
-				pdata = &data->phases[phase];
 			} else if (data->loop == -1 || thread_loop < data->loop - 1) {
 				/*
 				 * Phases are finished, but we might
@@ -833,8 +836,14 @@ void *thread_body(void *arg)
 				phase_loop = 0;
 				phase = 0;
 				thread_loop++;
-				pdata = &data->phases[0];
 			}
+
+			/*
+			 * If we reset phase_loop it means that phase now
+			 * points to a valid phase (next or first).
+			 */
+			if (!phase_loop)
+				pdata = &data->phases[phase];
 		}
 
 		log_idx++;
```
Otherwise we might point pdata to unknown memory after last iteration and phase_loop != pdata->loop check might segfault.

vingu-linaro: This seems to be quite complex compared to what was done before. IIUC, you don't use thread_loop anymore in the while because it will overflow and equals -1 at some point and the thread will stop, ins't it ?
I this case why not doing:
```
		phase_loop++;
		/* Reached the specified number of loops for this phase. */
		if (phase_loop == pdata->loop) {
			phase_loop = 0;

			phase++;
			/* Finished the last phase */
			if (phase = data->nphases) {
				phase = 0;

				thread_loop++;
				/* 
				 * a negative thread_loop means that it has overflowed and
				 * that data->loop == -1 which means infinite loop.
				 * Reset thread_loop to prevent stopping the thread
				 */
				if (thread_loop < 0)
					thread_loop = 0;
			}
			pdata = &data->phases[phase]
		}
```

vingu-linaro: and we use are back to the below for the main loop :
```
    while (continue_running && (thread_loop != data->loop)) {
```

vingu-linaro:
Last version looks good to me.
I'm going to merge it

PR #28, opened 2017-06-08
Fixes/loadwait perf

jlelli:
perf metric is broken when exec is multiple of 1s: it returns 0.
Fix the problem and propose to decrease perf metric granularity to reduce overflow probability.

### DIFF
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -570,7 +570,7 @@ finished to execute its events.
 
 You can log some metrics for each phase that is executed by a thread. These
 metrics are:
-- perf: number of loop that has been executed by run events
+- perf: fixed amount of work perfomed during a phase (c_duration/calibration)
 - run: time spent by the thread to execute the run events
 - period: duration to execute the complte phase
 - start/end : absolute start and end time of a phase. Same time base is used in
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -183,9 +183,15 @@ int calibrate_cpu_cycles(int clock)
 
 static inline unsigned long loadwait(unsigned long exec)
 {
-	unsigned long load_count;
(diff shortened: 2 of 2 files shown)


vingu-linaro:
1st commit is ok for me.
For the 2nd commit, 10k is really high for platform like hikey. The perf will be the same up to 1.5ms of runtime difference. This might be wrongly interpreted has freq scaling

vingu-linaro:
after thinking about this a bit more, may be we can compute a ratio to make perf value with microsecond granularity which would mean more than an hour before the overflow

jlelli:
you mean the following (as exec is in usec already) ?
```
index 68a9e23..e11ea79 100644
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -192,7 +192,7 @@ static inline unsigned long loadwait(unsigned long exec)
 	 * compute it here because both load_count and exec might be modified
 	 * below.
 	 */
-	perf = (exec * 1000)/(p_load * 10000);
+	perf = exec / p_load;
 
 	/*
 	 * If exec is still too big, let's run it in bursts
```

vingu-linaro:
In fact, can't we simply return the exec value which is already in us and represents the duration that we want at max capacity ?

jlelli:
But that's already c_duration, right? That is also why I was thinking we might just remove exec. :)

vingu-linaro:
Yes that's probably the simplest solution :-)

jlelli:
OK, but, since I don't think we can remove "perf" as it would most probably break other tools/scripts, would my last diff make still sense to you? Or we can just put always 0 and explain why..

vingu-linaro:
OK. so let use your last diff in this case

jlelli:
Argh! Need to updated the comment..

PR #29, opened 2017-06-13
rt-app: add per phase set priority event

vingu-linaro:
Add capability to change task's priority when starting a new phase

Only RT and CFS task are managed. Deadline could be done ins a 2nd step
as it will require more changes

# Conflicts:
#	src/rt-app.c

### DIFF
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -186,29 +186,14 @@ below).
 * instance : Integer. Define the number of threads that will be created with
 the properties of this thread object. Default value is 1.
 
-*** policy : String: Define the scheduling policy of the thread. default_policy
-is used if not defined
-
-* priority : Integer. Define the scheduling priority of the thread. The value
-must be aligned with the range allowed by the policy. The default priority is
-0 for sched_other class and 10 for other classes
-
-* dl-runtime : Integer: Define the runtime budget for deadline scheduling class.
-Default value is 0. The unit is usec.
-
-* dl-period : Integer. Define the period duration for deadline scheduling class.
-Default value is runtime. The unit is usec.
-
-* dl-deadline : Integer. Define the deadline parameter for deadline scheduling
-class. Default value is period. The unit is usec.
-
-* cpus: Array of Integer. Define the CPU affinity of the thread. Default
-value is all CPUs of the system. An example :  "cpus" : [0, 2, 3]
-
 * delay: Integer. Initial delay before a thread starts execution. The unit
 is usec.
 
-* phases: Object. The phases object described TBF
+* phases: Object. The phases object describes the behavior of the thread. This
+  behavior can be split in several distinct phases with their own events and
+  properties. See phase object parameter below.
+
+*** phase object ***
 
 If there is only 1 phase, the sequence of events can be directly put in the
 thread object instead of creating a phases object. As an example, the 2 objects
@@ -231,6 +216,56 @@ below are equals:
 		"sleep" : 10
 	}
 }
+
+*** thread and phase object properties ***
+
+Several properties can be set at thread and/or phase levels. All these
+properties are optional and default value will be used if nothing is defined.
+
+* loop: Integer. Define the number of times the parent object must be run.  The
+  parent object can be a phase or a thread. For phase object, the loop defines
+the number of time the phase will be executed before starting the next phase.
+For thread object, the loop defines the number of times that the complete
+"phases" object will be executed. The default value is -1 for thread object and
+1 for phases.
+
+*** scheduling policy
+
+The scheduling policy can be set at thread and phase levels. A default policy
+is set after the thread creation if nothing has been defined. Then, the
+scheduling properties can be changed at the beginning of each phase if it has
+been defined in the phase object. The behavior is quite similar to an event as
+a update happens only if one of the scheduling parameters has been set
+otherwise the scheduling paraleters will stay unchanged
+
+* policy : String. Define the scheduling policy of the thread. default_policy
+is used if not defined. Accepted value are:
+  - "SCHED_OTHER"
+  - "SCHED_RR"
+  - "SCHED_FIFO"
+  - "SCHED_DEADLINE"
+
+* priority : Integer. Define the scheduling priority of the thread. The value
+must be aligned with the range allowed by the policy. The default priority is
+0 for sched_other and sched_deadline class and 10 for rt classes
+
+* dl-runtime : Integer: Define the runtime budget for deadline scheduling class.
+Default value is 0. The unit is usec.  For backward compatibility, the
+key "runtime" will also be checked in thread object but special must be taken
+as this can conflict with runtime event.
+
+* dl-period : Integer. Define the period duration for deadline scheduling class.
+Default value is runtime. The unit is usec. For backward compatibility, the
+key "period" will also be checked in thread object.
+
+* dl-deadline : Integer. Define the deadline parameter for deadline scheduling
+class. Default value is period. The unit is usec.  For backward compatibility,
+the key "deadline" will also be checked in thread object.
+
+*** CPUs affinity
+
+* cpus: Array of Integer. Define the CPU affinity of the thread. Default
+value is all CPUs of the system. An example :  "cpus" : [0, 2, 3]
 
 "cpus" can be specified at task level or phase level. As an example, below
 creates two threads. One thread will run with affinity of CPU 2 and 3. The
@@ -268,13 +303,6 @@ inherit the affinity from the task level):
 		}
 	}
 }
-
-* loop: Integer. Define the number of times the parent object must be run.  The
-  parent object can be a phase or a thread. For phase object, the loop defines
-the number of time the phase will be executed before starting the next phase.
-For thread object, the loop defines the number of times that the complete
-"phases" object will be executed. The default value is -1 for thread object and
-1 for phases.
 
 *** events ***
 
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -590,6 +590,111 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
 	}
 }
 
+static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
+{
+	struct sched_param param;
+	policy_t policy;
+#ifdef DLSCHED
+	struct sched_attr dl_params;
+	pid_t tid;
+	unsigned int flags = 0;
+#endif
+	int ret;
+
+	if (sched_data == NULL)
+		return;
+
+	if (data->curr_sched_data == sched_data)
+		return;
+
+	if ((sched_data->policy == same) &&  (data->curr_sched_data)){
+		/* if policy not specificed, reuse previous policy */
+		sched_data->policy = data->curr_sched_data->policy;
+	}
+
+
+	switch (sched_data->policy)
+	{
+		case rr:
+		case fifo:
+			log_debug("[%d] setting scheduler %s priority %d", data->ind,
+					policy_to_string(sched_data->policy), sched_data->prio);
+			param.sched_priority = sched_data->prio;
+			ret = pthread_setschedparam(pthread_self(),
+					sched_data->policy,
+					&param);
+			if (ret != 0) {
+				log_critical("[%d] pthread_setschedparam"
+				     "returned %d", data->ind, ret);
+				errno = ret;
+				perror("pthread_setschedparam");
+				exit(EXIT_FAILURE);
+			}
+			break;
+
+		case other:
+			log_debug("[%d] setting scheduler %s priority %d", data->ind,
+					policy_to_string(sched_data->policy), sched_data->prio);
+
+
+			if (sched_data->prio > 19 || sched_data->prio < -20) {
+				log_critical("[%d] setpriority "
+					"%d nice invalid. "
+					"Valid between -20 and 19",
+					data->ind, sched_data->prio);
+				exit(EXIT_FAILURE);
+			}
+
+			ret = setpriority(PRIO_PROCESS, 0,
+					sched_data->prio);
+			if (ret != 0) {
+				log_critical("[%d] setpriority"
+				     "returned %d", data->ind, ret);
+				errno = ret;
+				perror("setpriority");
+				exit(EXIT_FAILURE);
+			}
+
+			data->lock_pages = 0; /* forced off for SCHED_OTHER */
+			break;
+
+#ifdef DLSCHED
+		case deadline:
+			log_debug("[%d] setting scheduler %s exec %lu, deadline %lu"
+					" period %lu", data->ind,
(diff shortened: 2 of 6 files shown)


vingu-linaro:
Juri, let me know if you think that I should send the patch on sched-tools <sched-tools@lists.linaro.org> for a wider discussion

jlelli:
I'd maybe send a reference to this pull request on the mailing list, saying that people are more than welcome to use GitHub for reviewing. What do you think?
I fear that doing the review with two media might complicate things, and GitHub worked OK for me for the review of my fixes, do you agree or would prefer to stick with mailing list emails?

vingu-linaro:
ok. let do like that.
I'm going to send email with the pull request reference on  sched-tools@lists.linaro.org and invite people interested by the topic to review it on github

src/rt-app.c
```diff
@@ -686,6 +670,40 @@ void *thread_body(void *arg)
...
 
+	data->current_prio = sched_prio;
+}
+
+void *thread_body(void *arg)
+{
+	thread_data_t *data = (thread_data_t*) arg;
+	phase_data_t *pdata;
+	log_data_t ldata;
+	struct sched_param param;
```
jlelli: I understand you would like to get deadline support in a second step, but don't we break deadline here?
IIUC, this is the sched_param param thread_body uses below to set deadline attributes, but set_thread_priority() operates on a different (local) sched_param param, so the one we actually use for the syscall might be uninitialized?

src/rt-app_types.h
```diff
@@ -151,6 +151,7 @@ typedef struct _phase_data_t {
 	int loop;
 	event_data_t *events;
 	int nbevents;
+	int sched_prio;
```
jlelli: What about we embed a sched_param instead, and use the sched_prio field only for the moment?
So that we don't have to change it again for DEADLINE support.

src/rt-app.c
vingu-linaro: oops, you are right. I wanted to postponed deadline case but i have done it wrong

src/rt-app_types.h
vingu-linaro: yes might be a good solution. I will give it a try

vingu-linaro:
Hi Juri, I have pushed the new version which now support are scheduling class

src/rt-app_parse_config.c
```diff
@@ -683,6 +683,74 @@ static void parse_cpuset_data(struct json_object *obj, cpuset_data_t *data)
...
+	else if (tmp_data.policy == other)
+		prior_def = DEFAULT_THREAD_NICE;
+	else
+		prior_def = DEFAULT_THREAD_PRIORITY;
+
+	tmp_data.prio = get_int_value_from(obj, "priority", TRUE, prior_def);
+
+	/* deadline params */
+	tmp_data.runtime = get_int_value_from(obj, "dl-runtime", TRUE, 0);
+	tmp_data.period = get_int_value_from(obj, "dl-period", TRUE, tmp_data.runtime);
```
jlelli: Not for this patch.. but these defaults don't make much sense I fear. :(

src/rt-app_parse_config.c
```diff
@@ -683,6 +683,74 @@ static void parse_cpuset_data(struct json_object *obj, cpuset_data_t *data)
...
+	tmp_data.deadline = get_int_value_from(obj, "dl-deadline", TRUE, tmp_data.period);
+
+
+	if (def_policy != -1) {
```
jlelli: Why only when we have a def_policy?
Maybe because we don't support legacy grammar inside phases?

src/rt-app.c
```diff
@@ -759,6 +772,15 @@ void *thread_body(void *arg)
 				NULL);
 	}
 
+	/* TODO find a better way to handle that constraint:
+	 * Set the task to SCHED_DEADLINE as far as possible touching its
+	 * budget as little as possible for the first iteration.
+	 */
+
+	/* Set scheduling policy and print pretty info on stdout */
+	log_notice("[%d] Starting with %s policy with priority %d", data->ind, policy_to_string(data->sched_data->policy), data->sched_data->prio);
```
jlelli: Nitpick: can we try to fit 80 columns? (here and above/below).
I don't think we respect the rule everywhere.. but for new changes.. :)

src/rt-app.c
```diff
@@ -759,6 +772,15 @@ void *thread_body(void *arg)
...
+	set_thread_priority(data, data->sched_data);
```
jlelli: Do we need to call this here? Considering that we call this again below when starting 1st phase.

doc/tutorial.txt
```diff
@@ -232,6 +217,45 @@ below are equals:
...
+* policy : String. Define the scheduling policy of the thread. default_policy
+is used if not defined. Accepted value are:
+  - "SCHED_OTHER"
+  - "SCHED_RR"
+  - "SCHED_FIFO"
+  - "SCHED_DEADLINE"
+
+* priority : Integer. Define the scheduling priority of the thread. The value
+must be aligned with the range allowed by the policy. The default priority is
+0 for sched_other class and 10 for rt classes
```
jlelli: Should be 0 for sched_deadline as well.

jlelli:
Do you think it's worth adding a json file example?

src/rt-app.c
```diff
@@ -666,26 +659,64 @@ void *thread_body(void *arg)
...
+			dl_params.sched_policy = SCHED_DEADLINE;
+			dl_params.sched_priority = 0;
+			dl_params.sched_runtime = sched_data->runtime;
+			dl_params.sched_deadline = sched_data->deadline; 
+			dl_params.sched_period = sched_data->period;
 
 			log_notice("[%d] period: %lu, exec: %lu, deadline: %lu",
-				data->ind, data->period, data->runtime, data->deadline);
-			break;
+				data->ind, sched_data->period, sched_data->runtime, sched_data->deadline);
```
jlelli: We don't print anything for the other classes, maybe we can remove this or we could change this to log_debug and add the same for the other cases as well?

src/rt-app.c
```diff
@@ -590,44 +590,39 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
...
 	unsigned int flags = 0;
-	int ret, phase, phase_loop, thread_loop, log_idx;
+#endif
+	int ret;
 
-	/* Set thread name */
-	ret = pthread_setname_np(pthread_self(), data->name);
-	if (ret !=  0) {
-		perror("pthread_setname_np thread name over 16 characters");
+	if (sched_data == NULL)
```
jlelli: I think we should fallback to the default policy settings if per phase sched_data is absent.

doc/tutorial.txt
```diff
@@ -232,6 +217,45 @@ below are equals:
...
+properties are optional and default value will be used if nothing is defined.
+
+* loop: Integer. Define the number of times the parent object must be run.  The
+  parent object can be a phase or a thread. For phase object, the loop defines
+the number of time the phase will be executed before starting the next phase.
+For thread object, the loop defines the number of times that the complete
+"phases" object will be executed. The default value is -1 for thread object and
+1 for phases.
+
+*** scheduling policy
```
jlelli: Do you think it might be better to create an Object for policy data?
It would contain policy name, priority and deadline parameters.
Similar to what we have for timers.

src/rt-app_parse_config.c
vingu-linaro: yes exactly.
anything new should follow dl-runtime/deadline/period, we keep runtime/deadline/period only for legacy support in thread object

src/rt-app.c
vingu-linaro: yes

src/rt-app.c
vingu-linaro: the next call can return without doing anything if there is anything define in phase object so this one can be the only call

doc/tutorial.txt
vingu-linaro: ok.

src/rt-app.c
vingu-linaro: I have move what we were doing before but i agree that it is not necessary or at least be moved in log_debug

src/rt-app.c
vingu-linaro: This is what is done for cpuset but i don't really like it for priority because with such behavior, you must add scheduler properties in every phase object even if you don't change compared to the previous phase. This behavior is more like other events in the phase for changing scheduler properties

doc/tutorial.txt
vingu-linaro: We can add the support of such policy object but we should keep compatibility with current behavior

src/rt-app_parse_config.c
vingu-linaro: which value should be put ? if runtime is not defined we can't do much with deadline scheduler. We can add a check when deadline scheduler is slected to make sure that runtiem/deadline/period are coherent

jlelli: right that's what I was thinking of (the syscall will for example fail with a zero runtime anyway for example)
but I can change that later no problem

src/rt-app_parse_config.c
jlelli: ok.. maybe add a comment or info in tutorial? (if not already there)

src/rt-app.c
jlelli: right
but even for tasks with no phases we will still execute the second call right (for the single phase that the task has)?
so I was thinking we might remove the call outside the main loop

src/rt-app.c
jlelli: ok again we can do it later no prob

src/rt-app.c
jlelli: mmm ok
maybe clarify it in tutorial?

doc/tutorial.txt
jlelli: true
ok we can do it later then.. if needed :)

src/rt-app.c
vingu-linaro: yes i will add explanation

src/rt-app.c
vingu-linaro: Yes we will call it twice in this case. We have 2 use cases that have problem:
1st config:
```
"threadA" : {
    "policy" : "SCHED_OTHER",
    "phases" : {
        "phase1" : {
            "run" : 5000,
            "timer" :  { "ref" : "tickB", "period" : 7777, }
        },
    },
}
```
The scheduling policy parameters are only at thread level but not at phase so it must be set before starting the loop. So we can't remove this call to set_thread_priority 

2nd config:
"threadA" : {
    "policy" : "SCHED_OTHER",
    "run" : 5000,
    "timer" :  { "ref" : "tickB", "period" : 7777, }
}
```
Because we don't have phase object, the scheduling policy parameters will be set at thread level and at phase level. the scheduling policy will be set twice.
But in fact, we don't have to allocate and set thread_data->sched_data in this case so the call to set_thread_priority above will return because sched_data is null. The change should be done in 
parse_thread_data() and data->sched_data = parse_sched_data(obj, opts->policy); should be called only if no "phases" object is defined. But in phase, we don't use default scheduling policy because it can be the same as previous phase so the config below will not work

```
"threadA" : {
    "run" : 5000,
    "timer" :  { "ref" : "tickB", "period" : 7777, }
}
```
I don't have simple solution for now. I will think a bit more about this

Comment (jlelli) [inline src/rt-app.c]: OK. But we seem to do the right thing for set_thread_affinity.
I wonder if we can take inspiration from that.

Comment (jlelli) [inline src/rt-app.c]: would this work?
```
        if (sched_data == NULL)
-               return;
+               sched_data = data->sched_data;
```

Comment (vingu-linaro) [inline src/rt-app.c]: This mean that we come back to thread config as soon as no policy as been defined in a phase which break the event based behavior

Comment (jlelli) [inline src/rt-app.c]: right.. another try below :-)
where I'm try to exploit the fact that curr_sched_data is initialized to NULL and only set the first time we call set_thread_priority()
```
index 856fe94..fe148eb 100644
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -616,10 +616,11 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
 	}
 }
 
-static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
+static void set_thread_priority(thread_data_t *data, sched_data_t *psched_data)
 {
 	struct sched_param param;
 	policy_t policy;
+	sched_data_t *sched_data = psched_data;
 #ifdef DLSCHED
 	struct sched_attr dl_params;
 	pid_t tid;
@@ -627,8 +628,12 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 #endif
 	int ret;
 
-	if (sched_data == NULL)
-		return;
+	if (sched_data == NULL) {
+		if (data->curr_sched_data == NULL)
+			sched_data = data->sched_data;
+		else
+			return;
+	}
 
 	if (data->curr_sched_data == sched_data)
 		return;
@@ -808,8 +813,10 @@ void *thread_body(void *arg)
 	 */
 
 	/* Set scheduling policy and print pretty info on stdout */
-	log_notice("[%d] Starting with %s policy with priority %d", data->ind, policy_to_string(data->sched_data->policy), data->sched_data->prio);
-	set_thread_priority(data, data->sched_data);
+	//log_notice("[%d] Starting with %s policy with priority %d",
+	//	   data->ind, policy_to_string(data->sched_data->policy),
+	//	   data->sched_data->prio);
+	//set_thread_priority(data, data->sched_data);
 
 	/*
 	 * phase        - index of current phase in data->phases array
```

Comment (vingu-linaro) [inline src/rt-app.c]: I'm afraid that this is not enough with config:
```
"threadA" : {
    "priority" : 0,
    "run" : 5000,
    "timer" :  { "ref" : "tickB", "period" : 7777, }
}
```
Because sched_data is not null but doesn't have default policy.

The change below should fix the problem:
```
@ -852,6 +852,14 @@ parse_thread_data(char *name, struct json_object *obj, int index,
 		data->phases = malloc(sizeof(phase_data_t) * data->nphases);
 		parse_thread_phase_data(obj,  &data->phases[0], opts, (long)data);
 
+		/* There is no "phases" object which means that thread and phase will
+		 * use same scheduling parameters. But thread object looks for default
+		 * value when parameters are not defined whereas phase doesn't.
+		 * We remove phase's scheduling policy which is a subset of thread's one
+		 */
+		free(data->phases[0].sched_data);
+		data->phases[0].sched_data = NULL;
+
 		/*
 		 * Get loop number:
 		 *
```

vingu-linaro:
Hi Juri,

I have pushed a new version of the patch which normally fixed all the issues we have discussed

Regards,
Vincent

jlelli:
Hi Vincent,

latest version looks all right. But it seems to conflict with master?
Could you see if you can solve merge conflicts and than merge the set?

Thanks.

vingu-linaro:
Hi Juri,

Ok. i'm going to solve it and merge the patchset.

Thanks,
Vincent

PR #32, opened 2017-07-06
Enhancing debug prints (v1)

jlelli:
I've been using Chris' ["Enhancing debug prints" PR](https://github.com/scheduler-tools/rt-app/pull/21) a lot lately.
So, I think it would be good to mainline his changes.
Talking with Chris he was fine with me posting his patches again (adding some modifications on top).

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -59,7 +59,7 @@ static pthread_barrier_t threads_barrier;
 static pthread_mutex_t joining_mutex;
 static pthread_mutex_t fork_mutex;
 
-static ftrace_data_t ft_data = {
+ftrace_data_t ft_data = {
 	.debugfs = "/sys/kernel/debug",
 	.marker_fd = -1,
 };
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -617,13 +617,15 @@ int run(int ind,
 	int nbevents = pdata->nbevents;
 	int i, lock = 0;
 	unsigned long perf = 0;
+	char type_name[RTAPP_RESOURCE_DESCR_LENGTH];
 
 	for (i = 0; i < nbevents; i++)
 	{
 		if (!continue_running && !lock)
 			return perf;
 
-		log_debug("[%d] runs events %d type %d ", ind, i, events[i].type);
+		log_debug("[%d] runs events %d type %s ", ind, i,
+			  ( resource_to_string(events[i].type, type_name), type_name) );
 		log_ftrace("[%d] executing %d", ind, i);
 		lock += run_event(&events[i], !continue_running, &perf,
 				  resources, t_first, ldata, ind);
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -366,7 +366,7 @@ static void memload(unsigned long count, struct _rtapp_iomem_buf *iomem)
 
 static int run_event(event_data_t *event, int dry_run,
 		unsigned long *perf, rtapp_resource_t *resources,
-		struct timespec *t_first, log_data_t *ldata)
+		struct timespec *t_first, log_data_t *ldata, int task_idx)
 {
 	rtapp_resource_t *rdata = &(resources[event->res]);
 	rtapp_resource_t *ddata = &(resources[event->dep]);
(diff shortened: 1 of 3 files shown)


src/rt-app.c
```diff
@@ -430,13 +430,15 @@ int run(int ind,
...
+	char type_name[64];
 
 	for (i = 0; i < nbevents; i++)
 	{
 		if (!continue_running && !lock)
 			return perf;
 
-		log_debug("[%d] runs events %d type %d ", ind, i, events[i].type);
+		log_debug("[%d] runs events %d type %d %s ", ind, i, events[i].type,
+			  ( resource_to_string(events[i].type, type_name), type_name) );
```
vingu-linaro: Is the event type number still needed now that you have added the name ?

src/rt-app.c
```diff
@@ -300,38 +304,49 @@ static int run_event(event_data_t *event, int dry_run,
 		pthread_mutex_unlock(&(rdata->res.barrier.m_obj));
 		break;
 	case rtapp_sig_and_wait:
-		log_debug("signal and wait %s", rdata->name);
+		log_debug("[%d] sync %s", task_idx, rdata->name);
+		log_debug("[%d] sync - signal %s", task_idx, rdata->name);
```
vingu-linaro: why do you use 2 lines ?
looks like these 2 lines display the same values ?

src/rt-app.c
```diff
@@ -300,38 +304,49 @@ static int run_event(event_data_t *event, int dry_run,
...
 	case rtapp_broadcast:
 		pthread_cond_broadcast(&(rdata->res.cond.obj));
 		break;
 	case rtapp_sleep:
 		{
 		struct timespec sleep = usec_to_timespec(event->duration);
-		log_debug("sleep %d ", event->duration);
-		nanosleep(&sleep, NULL);
+		log_debug("[%d] sleep %d (%ld.%09ld) start", task_idx, event->duration, sleep.tv_sec, sleep.tv_nsec);
+		while(1) {
```
vingu-linaro: sleep.tv_sec and  sleep.tv_nsec are redundant with duration

vingu-linaro: this new while(1) loop changes the behavior of sleep event. It's not about enhancing debug print

src/rt-app_utils.h
```diff
@@ -51,7 +51,7 @@ do {                                                                    \
     if (level <= LOG_LEVEL) {                                           \
         struct timespec t_now;                                          \
         clock_gettime(CLOCK_MONOTONIC, &t_now);                         \
-        fprintf(where, LOG_PREFIX level_pfx " %06ld.%09ld " msg "\n", t_now.tv_sec, t_now.tv_nsec, ##args);		\
+        fprintf(where, LOG_PREFIX "[%06ld.%09ld] " level_pfx msg "\n", t_now.tv_sec, t_now.tv_nsec, ##args);		\
```
vingu-linaro: Why do you want to change the log format ?
Would be good to give the reason in commit message

src/rt-app.c
credp: IIRC, the values of sleep are updated if a signal stops the sleep - this lets you know how much time passed before the code was interrupted.

credp: I agree, this is a change - we should probably review that as a separate patch. The intent is for signals not to interrupt the desired sleep time, but this might not be desirable or necessary.

src/rt-app.c
credp: IIRC, the problems I was debugging in playing back complex rt-app scripts were made easier by knowing what the event index was as well as the type, so I would quite like to keep if possible.

src/rt-app.c
credp: Fixed this

src/rt-app.c
credp: This is now a separate commit.

credp: change separated out.

src/rt-app_utils.h
credp: This is a change Juri made - @jlelli can you remember why you wanted to do this? I just rolled it up as part of the original commit in the latest rebase.

jlelli: > can you remember why you wanted to do this?

Not really, it was a while ago. :-)

Go ahead and remove this part if you like. Thanks!

src/rt-app_utils.c
```diff
@@ -266,8 +266,29 @@ resource_to_string(resource_t resource, char *resource_name)
 			break;
 		case rtapp_barrier:
 			strcpy(resource_name, "barrier");
```
credp: There's a missing break here. Will fix and repush it..

src/rt-app.c
credp: I'm going to check this is actually ever meaningful and take it out if not.

credp: This is indeed bogus

src/rt-app_utils.h
credp: Done - I've kept the timestamp and moved the level_pfx back where it should be.

src/rt-app.c
credp: I've moved this and a couple of other things into a separate branch on my own fork https://github.com/credp/rt-app/tree/features/cleanups

credp:
Yeah, that looks fine to me - all the useful information is there. I'll close this one.

PR #33, opened 2017-07-14
Pipeline latency

jlelli:
Hi,
I've been thinking it would be helpful to have a metric to measure end-to-end latency of pipelines of tasks (to compare against energy consumption).
The idea is to propose rt-app as a tool for creating mobile like type of workloads during the [Scheduler Workloads uconf](http://wiki.linuxplumbersconf.org/2017:scheduler_workloads) at LPC (if accepted).
Still highly WIP, I'm sure it doesn't always work reliably, but get some early feedback would be much appreciated.


src/rt-app.c
```diff
@@ -286,18 +286,29 @@ static int run_event(event_data_t *event, int dry_run,
...
-		if (rdata->res.barrier.waiting == 0) {
-			/* everyone is already waiting, signal */
-			pthread_cond_broadcast(&(rdata->res.barrier.c_obj));
-		} else {
-			/* not everyone is waiting, mark then wait */
-			rdata->res.barrier.waiting -= 1;
-			pthread_cond_wait(&(rdata->res.barrier.c_obj), &(rdata->res.barrier.m_obj));
-			rdata->res.barrier.waiting += 1;
+		{
+			struct timespec t_delta;
```
credp: This seems a bit nitpicky, so feel free to ignore it, but you could put t_delta inside the if condition to avoid the additional {} indent.

jlelli:
Hey, I didn't unfortunately.
Let me see if I'll find any time to work on this.
Someone else can pick it up as well of course.

PR #42, opened 2017-12-06
Return simple help and version

balsini:
Since it often happens to distribute already (cross)compiled rt-app
executables, it is useful to add a way to check the release version of
the software.
This patch adds the software version and build information to the
executable, that can be accessed with the arguments -v or --version
with the following syntax:

[PACKAGE] [git describe --tags HEAD] ([BUILD DATE])

$ rt-app -v|--version
rt-app v1.0-29-g6f10c0c (2017-12-06 16:39:38 GMT)

Moreover, a simple help is returned when using the arguments -h or
--help.

### DIFF
--- a/src/rt-app_args.c
+++ b/src/rt-app_args.c
@@ -23,36 +23,79 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include <stdio.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <getopt.h>
 
 #include "rt-app_parse_config.h"
 
+char help_usage[] = \
+"Usage: rt-app <taskset.json>\n"
+"Try 'rt-app --help' for more information.\n";
+
+char help_full[] = \
+"Usage:\n"
+"      rt-app <taskset.json>\n"
+"      cat taskset.json | rt-app -\n\n"
+"taskset.json is a json file describing the workload that will be generated"
+"by rt-app.\n\n"
+"In the first example, the json file is opened and parsed by rt-app.\n"
+"In the second example, rt-app reads the workload description in json format\n"
+"through the standard input.\n\n"
+"Miscellaneous:\n"
+"  -v, --version             display version information and exit\n"
+"  -h, --help                display this help text and exit\n";
+
 void
-usage (const char* msg, int ex_code)
+usage(const char* msg, int ex_code)
 {
-	printf("usage:\n"
-	       "rt-app <taskset.json>\n");
+	printf("%s", help_usage);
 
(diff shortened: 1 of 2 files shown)


src/rt-app_args.c
```diff
@@ -23,36 +23,79 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include <stdio.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <getopt.h>
 
 #include "rt-app_parse_config.h"
 
+char help_usage[] = \
+"Usage: rt-app <taskset.json>\n"
```
derkling: What about reporting the version also here, before this short usage, to generate an output similar to this:

```bash
$> ./src/rt-app
rt-app v1.0-29-gd04f322 (2017-12-08 10:37:49 GMT)

Usage: rt-app <taskset.json>

Try 'rt-app --help' for more information.

$>
```

version should probably be added at L87

src/rt-app_args.c
```diff
@@ -23,36 +23,79 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
...
-}
+	if (optind >= argc)
+		usage(NULL, EXIT_INV_COMMANDLINE);
 
+	if (stat(argv[optind], &config_file_stat) == 0)
+		parse_config(argv[optind], opts);
+	else if (strcmp(argv[optind], "-") == 0)
+		parse_config_stdin(opts);
+	else
+		usage(NULL, EXIT_FAILURE);
```
derkling: Maybe the error it's more readable if you add an empty line before the usage notes.

src/rt-app_args.c
balsini: I decided not to add it because many popular tools provide the version only when asked.

src/rt-app_args.c
balsini: I agree, but this choice has been done to make it as similar as the popular tools:
```
$ tar
tar: You must specify one of the '-Acdtrux', '--delete' or '--test-label' options
Try 'tar --help' or 'tar --usage' for more information.

$ src/rt-app 
Usage: rt-app <taskset.json>
Try 'rt-app --help' for more information.
```
Does it make sense?

src/rt-app_args.c
vingu-linaro: I agree with balsini, that we don't want to pollute rt-app output unnecessarily with version each time it's executed

src/rt-app_args.c
derkling: ok, fine with me! ;-)

PR #44, opened 2017-12-07
Bugfix for using logging macros code blocks without brackets

balsini:
The semicolon after the "while(0)" in defining the language trick:

do { \
    rtapp_log_to(stderr, LOG_LEVEL_ERROR, "<error> ", msg, ##args); \
} while (0);

is not necessary, and represents an error when using logging macros like
log_debug() in the following scenarios:

if (condition)
	log_debug("condition true");
else
	log_debug("condition false");


derkling:
LGTM

PR #45, opened 2017-12-07
Bugfix on handling wrong dl parameters

balsini:
In the following trivial workload example, where "exec" is used instead of
"runtime", rt-app crashes with a segmentation fault without notifying
the user of what was wrong.

{
  "global": {
    "duration": 10,
    "default_policy": "SCHED_DEADLINE"
  },
  "tasks": {
    "Task_0": {
      "deadline": 3000000,
      "period": 3000000,
      "exec": 2900000
    }
  }
}

This patch:
 - solves the problem by imposing the correctness of the dl parameters
   or notifying the error and exiting.
 - Refactors the code that parses dl parameters.
 - Added duplicates of logging printers for dl tasks, printing runtime,
   deadline and period instead of priority.

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -785,9 +785,20 @@ void *thread_body(void *arg)
 	 */
 
 	/* Set scheduling policy and print pretty info on stdout */
-	log_notice("[%d] Starting with %s policy with priority %d",
-			data->ind, policy_to_string(data->sched_data->policy),
-			data->sched_data->prio);
+	if (data->sched_data->policy == deadline) {
+		log_notice("[%d] Starting with %s policy with runtime %ld, deadline %ld, period %ld",
+			   data->ind,
+			   policy_to_string(data->sched_data->policy),
+			   data->sched_data->runtime,
+			   data->sched_data->deadline,
+			   data->sched_data->period);
+	} else {
+		log_notice("[%d] Starting with %s policy with priority %d",
+			   data->ind,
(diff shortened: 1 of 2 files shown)


src/rt-app.c
```diff
@@ -785,9 +785,20 @@ void *thread_body(void *arg)
...
+			   data->sched_data->runtime,
+			   data->sched_data->deadline,
+			   data->sched_data->period);
+	} else {
+		log_notice("[%d] Starting with %s policy with priority %d",
+			   data->ind,
+			   policy_to_string(data->sched_data->policy),
+			   data->sched_data->prio);
+	}
+
```
vingu-linaro: all these parameters duplicated with what is already print by set_thread_priority
The only reasonable modification could be to remove the priority information from the starting log

vingu-linaro:
The crash is not link to deadline parameters but to the fact that we don't have events to run for the thread.
The proper fix for this crash is https://github.com/scheduler-tools/rt-app/commits/fix-no-events-thread
Then, deadline, period and runtime are deprecated, dl-period, dl-deadline and dl-runtime must be used instead because runtime is also a event for running a specified amount of time whatever the frequency.
Then, we try to make all parameters optional as much as possible in order to keep the json fill small. That's why no scheduler class properties are mandatory but only optional. dl-deadline will be equal to dl-period if not defined, dl-period will be equal to dl-runtime if not-defined and dl-runtime equals if not defined.
We don't need duplicated information at different log level

PR #49, opened 2018-01-02
Add taskgroup support for SCHED_OTHER tasks

deggeman:
The implementation is based on libcgroup. It can be optionally
configured by specifying --with-cgroup.

Building instruction and an example are given in doc/tutorial.txt.

For now an rt-app binary w/ taskgroup support only allows SCHED_NORMAL
policy.

It can run on a kernel w/o cpu controller support in case there are
no taskgroups other than a root ("/") or an empty taskgroup ("")
configured.

An rt-app binary w/o taskgroup support ignores any taskgroup related
name/value pairs.

Tested on x86_64 and arm64.

Known issue: The build shows some warnings that some of the functions
in libcgroup.a (e.g. cgroup_change_cgroup_flags()) are calling
functions (e.g. getgrgid()) which require at run-time the same
version of shared libraries from the glibc version used for linking.

Signed-off-by: Dietmar Eggemann <dietmar.eggemann@arm.com>

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -37,6 +37,50 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include "rt-app_utils.h"
 #include "rt-app_args.h"
 
+#ifdef HAVE_LIBCGROUP
+/*
+ * Save and restore the following rt-app defines from src/config.h. libcgroup
+ * defines its own set and its public header libcgroup.h includes libcgroup's
+ * config.h.
+ */
+#define _VERSION VERSION
+#define _PACKAGE PACKAGE
+#define _PACKAGE_NAME PACKAGE_NAME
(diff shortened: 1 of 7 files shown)


mdigiorgio:
This is something useful for creating hierarchies of tasks to test PELT signals for taskgroups. Looking forward to seeing it merged.

src/rt-app.c
```diff
@@ -539,6 +583,223 @@ static void create_cpuset_str(cpuset_data_t *cpu_data)
...
+			int n = strcmp(new->name, tg->name);
+
+			if (n < 0) {
+				TAILQ_INSERT_BEFORE(tg, new, taskgroups);
+				return;
+			}
+
+			/* Do not add taskgroup twice */
+			if (!n)
+				return;
```
vingu-linaro: you have to free the malloc in this case

PR #51, opened 2018-05-28
fix sched_other policy setting

vingu-linaro:
Use pthread_setschedparam instead of setpriority for setting task into
CFS class. setpriority only changes the CFS priority but not the scheduling
class.
When a task has been set to fifo class, it can't move back to other class

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -648,13 +648,15 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 				exit(EXIT_FAILURE);
 			}
 
-			ret = setpriority(PRIO_PROCESS, 0,
-					sched_data->prio);
+			param.sched_priority = sched_data->prio;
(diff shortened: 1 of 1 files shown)


src/rt-app.c
```diff
@@ -648,8 +648,10 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
...
 			}
 
-			ret = setpriority(PRIO_PROCESS, 0,
-					sched_data->prio);
+			param.sched_priority = sched_data->prio;
+			ret = pthread_setschedparam(pthread_self(),
+					sched_data->policy,
+					&param);
 			if (ret != 0) {
 				log_critical("[%d] setpriority"
```
jlelli: Nitpick: we should change the error message as well.

jlelli: Nitpick: we should change the error message as well.

vingu-linaro: yes you're right

PR #52, opened 2018-06-13
fix sched_other policy setting

vingu-linaro:
pthread_setschedparam can only change policy to SCHED_OTHER but can not set
nice priority (It can only set rt priority). So we must keep calling setpriority to set nice value


jlelli:
Ah indeed, thanks!

PR #53, opened 2018-06-29
Fork support

qais-yousef:
To be able to test scheduler behaviour when forking a new thread, we need to be able to fork half way through the experiment.

This pull request adds a new 'fork' event to fork a new thread at a selected point of time rather just at rt-app startup. The new forked threads doesn't accept phases for simplicity. And there's a hard limit of 1024 forks to keep the resource allocation static at rt-app start rather than dynamically allocate at runtime which could 'interfere' with the experiment.

The pull request also fixes a typo in tutorial.txt which states that the default value of lock_pages is false but it is actually true.

I still need to update tutorial.txt to include the new fork event, but first want to make sure the approach is acceptable.

Thanks.

### DIFF
--- a/doc/examples/tutorial/example9.json
+++ b/doc/examples/tutorial/example9.json
@@ -1,0 +1,56 @@
+{
+	/*
+	 * A simple example on using fork event.
+	 *
+	 * thread3 will fork thread1 and thread2 in each of its phases.
+	 * Note that thread2 will NOT be instantiated at start time and
+	 * will only run when thread3 forks it. Unlike thread1 which will
+	 * run at the start - since default value of instance property is
+	 * 1 - and will run again when thread3 forks ie: twice in total.
+	 */
+	"tasks" : {
+		"thread1" : {
+			"phases" : {
+				"phase1" : {
+					"run" : 10000,
+					"sleep" : 10000
+				}
+			}
+		},
+		"thread2" : {
+			"instance": 0,
+			"phases" : {
+				"phase1" : {
+					"run" : 20000,
+					"sleep" : 20000
+				}
+			}
+		},
+		"thread3" : {
+			"phases" : {
+				"phase1" : {
+					"fork": "thread1",
+					"run" : 10000,
+					"sleep" : 10000
+				},
+				"phase2" : {
+					"fork": "thread2",
+					"run" : 20000,
+					"sleep" : 20000
+				}
+			},
+			"loop": 1
+		}
+	},
+	"global" : {
+		"duration" : 2,
+		"calibration" : "CPU0",
+		"default_policy" : "SCHED_OTHER",
+		"pi_enabled" : false,
+		"lock_pages" : false,
+		"logdir" : "./",
+		"log_basename" : "rt-app1",
+		"ftrace" : false,
+		"gnuplot" : false
+	}
+}
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -136,7 +136,8 @@ several threads can be created with the same object (see instance parameter
 below).
 
 * instance : Integer. Define the number of threads that will be created with
-the properties of this thread object. Default value is 1.
+the properties of this thread object. Default value is 1. A value of 0 will
+create the structures of the task but will not create a pthread.
 
 * delay: Integer. Initial delay before a thread starts execution. The unit
 is usec.
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -489,6 +489,10 @@ taskB ...-------------------|resume taskA|run 10    |--------------------
 * yield: String. Calls pthread_yield(), freeing the CPU for other tasks. This has a
 special meaning for SCHED_DEADLINE tasks. String can be empty.
 
+* fork: String. Instead of creating all tasks at start time, you can fork one
+at any point of time using this event. The name of the forked task must match
+one of the defined tasks.
+
 **** Trace and Log ****
 
 Some traces and log hooks have been added to ease the debug and monitor various
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -71,7 +71,7 @@ static thread_data_t *find_thread_data(const char *name, rtapp_options_t *opts)
 {
 	int i;
 
-	for (i = 0; i < opts->nthreads; i++) {
+	for (i = 0; i < opts->num_tasks; i++) {
 		thread_data_t *tdata = &opts->threads_data[i];
 		if (!strcmp(tdata->name, name))
 			return tdata;
@@ -106,6 +106,64 @@ static void thread_data_set_unique_name(thread_data_t *tdata, int nforks)
 		 * and we'll know */
 		tdata->name = strdup(tdata->name);
 	}
+}
+
+/*
+ * Create a new pthread using the info provided in the passed @td thread_data_t.
+ *
+ * @td:		The thread data of the task to create - we copy it and leave
+ *		it unmodified for other fork events to use to create additional
+ *		threads based on the same tdata.
+ *
+ * @index:	Index of the task to create.
+ *
+ * @forked:	Whether this task is a craeted from fork event or at
+ *		application startup.
+ *
+ * @nforks:	If this is a forked task, we use nforks to give it a unique name.
+ *
+ * Returns 0 on success or -1 on failure.
+ */
+static int create_thread(const thread_data_t *td, int index, int forked, int nforks)
+{
+	thread_data_t *tdata;
+
+	if (!td) {
+		log_error("Failed to create new thread, passed NULL thread_data_t: %s", td->name);
+		return -1;
+	}
+
+	tdata = malloc(sizeof(thread_data_t));
+	if (!tdata) {
+		log_error("Failed to duplicate thread data: %s", td->name);
+		return -1;
+	}
+
+	/*
+	 * We have one tdata created at config parse, but we
+	 * might spawn multiple threads if we were running in
+	 * a loop, so ensure we duplicate the tdata before
+	 * creating each thread, and we should free it at the
+	 * end of the thread_body()
+	 */
+	memcpy(tdata, td, sizeof(*tdata));
+
+	/* Mark this thread as forked */
+	tdata->forked = forked;
+	/* update the index value */
+	tdata->ind = index;
+
+	/* Make sure each (forked) thread has a unique name */
+	thread_data_set_unique_name(tdata, nforks);
+
+	setup_thread_logging(tdata);
+
+	if (pthread_create(&threads[index], NULL, thread_body, (void*) tdata)) {
+		perror("Failed to create a thread");
+		return -1;
+	}
+
+	return 0;
 }
 
 /*
@@ -490,9 +548,8 @@ static int run_event(event_data_t *event, int dry_run,
 			pthread_mutex_lock(&fork_mutex);
 			int new_thread_index = nthreads++;
 			threads = realloc(threads, nthreads * sizeof(*threads));
-			thread_data_t *tdata = malloc(sizeof(thread_data_t));
 
-			if (!threads || !tdata) {
+			if (!threads) {
 				log_error("Failed to allocate memory for a new fork: %s", rdata->res.fork.tdata->name);
 				pthread_mutex_unlock(&fork_mutex);
 				exit(EXIT_FAILURE);
@@ -513,27 +570,8 @@ static int run_event(event_data_t *event, int dry_run,
 				rdata->res.fork.tdata = find_thread_data(rdata->res.fork.ref, &opts);
 			}
 
-			/*
-			 * We have one tdata created at config parse, but we
-			 * might spawn multiple threads if we were running in
-			 * a loop, so ensure we duplicate the tdata before
-			 * creating each thread, and we should free it at the
-			 * end of the thread_body()
-			 */
-			memcpy(tdata, rdata->res.fork.tdata, sizeof(*tdata));
-
-			/* Mark this thread as forked */
-			tdata->forked = 1;
-			/* update the index value */
-			tdata->ind = new_thread_index;
-
-			/* Make sure each forked thread has a unique name */
-			thread_data_set_unique_name(tdata, rdata->res.fork.nforks++);
-
-			setup_thread_logging(tdata);
-
-			if (pthread_create(&threads[new_thread_index], NULL, thread_body, (void*) tdata)) {
-				perror("Failed to create a forked thread");
+			int ret = create_thread(rdata->res.fork.tdata, new_thread_index, 1, rdata->res.fork.nforks++);
+			if (ret) {
 				pthread_mutex_unlock(&fork_mutex);
 				exit(EXIT_FAILURE);
 			}
@@ -1127,7 +1165,6 @@ int main(int argc, char* argv[])
 {
 	FILE *gnuplot_script = NULL;
 	int i, res, nresources;
-	thread_data_t *tdata;
 	rtapp_resource_t *rdata;
 	char tmp[PATH_LENGTH];
 	static cpu_set_t orig_set;
@@ -1307,7 +1344,8 @@ int main(int argc, char* argv[])
 	clock_gettime(CLOCK_MONOTONIC, &t_start);
 
 	/* Start the use case */
-	for (i = 0; i < nthreads; i++) {
+	int ind = 0;
+	for (i = 0; i < opts.num_tasks; i++) {
 		/*
 		 * Duplicate thread data so that we can safely copy the
 		 * original thread data when forking.
@@ -1320,22 +1358,20 @@ int main(int argc, char* argv[])
 		 * are intact and duplicate them before forking or when we run
 		 * for the first time as in here.
 		 */
-		tdata = malloc(sizeof(thread_data_t));
-		if (!tdata) {
-			perror("Failed to duplicate thread data");
-			exit(EXIT_FAILURE);
-		}
-		memcpy(tdata, &opts.threads_data[i], sizeof(*tdata));
-		/* Make sure to update the name with something unique */
-		thread_data_set_unique_name(tdata, 0);
+		thread_data_t *tdata_orig = &opts.threads_data[i];
+		int j;
 
-		setup_thread_logging(tdata);
-
-		if (pthread_create(&threads[i],
-				  NULL,
-				  thread_body,
-				  (void*) tdata))
+		if (tdata_orig->num_instances < 0) {
+			log_error("Invalid num_instances value: %d", tdata_orig->num_instances);
 			goto exit_err;
+		}
+
+		for (j = 0; j < tdata_orig->num_instances; j++) {
+			int ret = create_thread(tdata_orig, ind++, 0, 0);
+			if (ret) {
+				goto exit_err;
+			}
+		}
 	}
 	running_threads = nthreads;
 
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -46,13 +46,59 @@ static volatile sig_atomic_t running_threads;
 static int p_load;
 rtapp_options_t opts;
 static struct timespec t_zero;
+static struct timespec t_start;
 static pthread_barrier_t threads_barrier;
 static pthread_mutex_t joining_mutex;
+static pthread_mutex_t fork_mutex;
 
 static ftrace_data_t ft_data = {
 	.debugfs = "/sys/kernel/debug",
 	.marker_fd = -1,
 };
+
+void *thread_body(void *arg);
+void setup_thread_logging(thread_data_t *tdata);
+
+static thread_data_t *find_thread_data(const char *name, rtapp_options_t *opts)
+{
+	int i;
+
+	for (i = 0; i < opts->nthreads; i++) {
+		thread_data_t *tdata = &opts->threads_data[i];
+		if (!strcmp(tdata->name, name))
+			return tdata;
+	}
+
+	log_error("Can't fork unknown task %s", name);
+	exit(EXIT_FAILURE);
+}
+
+/*
+ * Give each created thread a unique name based on tdata->ind. If the thread is
+ * forked we track each fork by a unique number that is incremented
+ * independently for each fork-event.
+ */
+static void thread_data_set_unique_name(thread_data_t *tdata, int nforks)
+{
+	int string_size = strlen(tdata->name) + 1 /* NULL */ + 10 /* postfix */;
+	char *unique_name = malloc(string_size);
+	if (unique_name) {
+		if (tdata->forked) {
+			snprintf(unique_name, string_size, "%s-%d-%04d", tdata->name, tdata->ind, nforks);
+		} else {
+			snprintf(unique_name, string_size, "%s-%d", tdata->name, tdata->ind);
+		}
+		/*
+		 * no need to free tdata->name since tdata is a copy itself
+		 */
+		tdata->name = unique_name;
+	} else {
+		log_error("Failed to create a unique name for forked thread %s", tdata->name);
+		/* if strdup() fails (again) at least we'll get a null asigned
+		 * and we'll know */
+		tdata->name = strdup(tdata->name);
+	}
+}
 
 /*
  * Function: to do some useless operation.
@@ -415,6 +461,70 @@ static int run_event(event_data_t *event, int dry_run,
 			pthread_yield();
 		}
 		break;
+	case rtapp_fork:
+		{
+			log_debug("fork %s", rdata->res.fork.ref);
+
+			/*
+			 * If multiple threads race to fork, we must ensure
+			 * each one sees a unique index. Hence the lock.
+			 */
+			pthread_mutex_lock(&fork_mutex);
+			int new_thread_index = nthreads++;
+			threads = realloc(threads, nthreads * sizeof(*threads));
+			thread_data_t *tdata = malloc(sizeof(thread_data_t));
+
+			if (!threads || !tdata) {
+				log_error("Failed to allocate memory for a new fork: %s", rdata->res.fork.tdata->name);
+				pthread_mutex_unlock(&fork_mutex);
+				exit(EXIT_FAILURE);
+			}
+
+			/*
+			 * Find the thread data associated with this fork and
+			 * store it in tdata; this needs to be done once if the
+			 * fork is done within a loop.
+			 *
+			 * Note that we can't search for the reference at parse
+			 * time because we are not guaranteed that the task
+			 * that is referenced was already parsed; it'll depend
+			 * greatly on the ordering within the defined json
+			 * file.
+			 */
+			if (!rdata->res.fork.tdata) {
+				rdata->res.fork.tdata = find_thread_data(rdata->res.fork.ref, &opts);
+			}
+
+			/*
+			 * We have one tdata created at config parse, but we
+			 * might spawn multiple threads if we were running in
+			 * a loop, so ensure we duplicate the tdata before
+			 * creating each thread, and we should free it at the
+			 * end of the thread_body()
+			 */
+			memcpy(tdata, rdata->res.fork.tdata, sizeof(*tdata));
+
+			/* Mark this thread as forked */
+			tdata->forked = 1;
+			/* update the index value */
+			tdata->ind = new_thread_index;
+
+			/* Make sure each forked thread has a unique name */
+			thread_data_set_unique_name(tdata, rdata->res.fork.nforks++);
+
+			setup_thread_logging(tdata);
+
+			if (pthread_create(&threads[new_thread_index], NULL, thread_body, (void*) tdata)) {
+				perror("Failed to create a forked thread");
+				pthread_mutex_unlock(&fork_mutex);
+				exit(EXIT_FAILURE);
+			}
+
+			running_threads = nthreads;
+
+			pthread_mutex_unlock(&fork_mutex);
+		}
+		break;
 	default:
 		break;
 	}
@@ -487,7 +597,16 @@ static void __shutdown(bool force_terminate)
 	if (pthread_mutex_trylock(&joining_mutex))
 		return;
 
-	/* wait up all waiting threads */
+	/*
+	 * Wait for all threads to terminate.
+	 *
+	 * pthread_join() will block until the process has terminated. If any
+	 * thread forks any processes then it'll update running_threads before
+	 * it returns and since running_threads is volatile the for() loop is
+	 * guaranteed to check against the updated value in each iteration.
+	 * Hence we are guaranteed to wait for all forked threads beside the
+	 * originally created ones at startup time.
+	 */
 	for (i = 0; i < running_threads; i++)
 	{
 		int ret = pthread_join(threads[i], NULL);
@@ -805,7 +924,9 @@ void *thread_body(void *arg)
 				data->ind);
 	}
 
-	pthread_barrier_wait(&threads_barrier);
+	if (!data->forked)
+		pthread_barrier_wait(&threads_barrier);
+
 	t_first = t_zero;
 
 	log_notice("[%d] starting thread ...\n", data->ind);
@@ -954,13 +1075,38 @@ void *thread_body(void *arg)
 	if (opts.logsize)
 		fclose(data->log_handler);
 
+	/* clean up tdata if this was a forked thread */
+	free(data->name);
+	free(data);
+
 	pthread_exit(NULL);
 }
 
+void setup_thread_logging(thread_data_t *tdata)
+{
+	char tmp[PATH_LENGTH];
+
+	tdata->duration = opts.duration;
+	tdata->main_app_start = t_start;
+	tdata->lock_pages = opts.lock_pages;
+
+	if (opts.logdir) {
+		snprintf(tmp, PATH_LENGTH, "%s/%s-%s.log",
+			 opts.logdir,
+			 opts.logbasename,
+			 tdata->name);
+		tdata->log_handler = fopen(tmp, "w");
+		if (!tdata->log_handler) {
+			log_error("Cannot open logfile %s", tmp);
+			exit(EXIT_FAILURE);
+		}
+	} else {
+		tdata->log_handler = stdout;
+	}
+}
 
 int main(int argc, char* argv[])
 {
-	struct timespec t_start;
 	FILE *gnuplot_script = NULL;
 	int i, res, nresources;
 	thread_data_t *tdata;
@@ -975,6 +1121,7 @@ int main(int argc, char* argv[])
 	threads = malloc(nthreads * sizeof(pthread_t));
 	pthread_barrier_init(&threads_barrier, NULL, nthreads);
 	pthread_mutex_init(&joining_mutex, NULL);
+	pthread_mutex_init(&fork_mutex, NULL);
 
 	/* install a signal handler for proper shutdown */
 	signal(SIGQUIT, shutdown);
@@ -1020,32 +1167,6 @@ int main(int argc, char* argv[])
 
 	/* Take the beginning time for everything */
 	clock_gettime(CLOCK_MONOTONIC, &t_start);
-
-	if (opts.logsize) {
-	/* Prepare log file of each thread before starting the use case */
-		for (i = 0; i < nthreads; i++) {
-			tdata = &opts.threads_data[i];
-
-			tdata->duration = opts.duration;
-			tdata->main_app_start = t_start;
-			tdata->lock_pages = opts.lock_pages;
-
-			if (opts.logdir) {
-				snprintf(tmp, PATH_LENGTH, "%s/%s-%s-%d.log",
-					 opts.logdir,
-					 opts.logbasename,
-					 tdata->name,
-					 tdata->ind);
-				tdata->log_handler = fopen(tmp, "w");
-				if (!tdata->log_handler) {
-					log_error("Cannot open logfile %s", tmp);
-					exit(EXIT_FAILURE);
-				}
-			} else {
-				tdata->log_handler = stdout;
-			}
-		}
-	}
 
 	/* Prepare gnuplot files before starting the use case */
 	if (opts.logdir && opts.gnuplot) {
@@ -1169,7 +1290,28 @@ int main(int argc, char* argv[])
 
 	/* Start the use case */
 	for (i = 0; i < nthreads; i++) {
-		tdata = &opts.threads_data[i];
+		/*
+		 * Duplicate thread data so that we can safely copy the
+		 * original thread data when forking.
+		 *
+		 * If we don't do that and try to fork a running thread, we
+		 * might have partially modified content of thread data since
+		 * thread_body() calls functions like set_thread_affinity()
+		 * that modifies thread data at runtime. Rather than introduce
+		 * complex locking ensure that the original parsed thread_data
+		 * are intact and duplicate them before forking or when we run
+		 * for the first time as in here.
+		 */
+		tdata = malloc(sizeof(thread_data_t));
+		if (!tdata) {
+			perror("Failed to duplicate thread data");
+			exit(EXIT_FAILURE);
+		}
+		memcpy(tdata, &opts.threads_data[i], sizeof(*tdata));
+		/* Make sure to update the name with something unique */
+		thread_data_set_unique_name(tdata, 0);
+
+		setup_thread_logging(tdata);
 
 		if (pthread_create(&threads[i],
 				  NULL,
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -38,6 +38,14 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include "rt-app_utils.h"
 #include "rt-app_args.h"
 
+/*
+ * To prevent infinite loops in fork bombs, we will limit the number of
+ * permissible forks before we exit on error.
+ *
+ * This limit is per forking task/thread NOT the aggregated number of forks.
+ */
+#define FORKS_LIMIT		1024
+
 static int errno;
 static volatile sig_atomic_t continue_running;
 static pthread_t *threads;
@@ -464,6 +472,16 @@ static int run_event(event_data_t *event, int dry_run,
 	case rtapp_fork:
 		{
 			log_debug("fork %s", rdata->res.fork.ref);
+
+			/*
+			 * Check if the current thread reached its limit of
+			 * number of allowable forks.
+			 * We enforce a limit to prevent infinite loops.
+			 */
+			if (rdata->res.fork.nforks >= FORKS_LIMIT) {
+				log_error("%s reached its fork limit (%d)", rdata->res.fork.tdata->name, FORKS_LIMIT);
+				exit(EXIT_FAILURE);
+			}
 
 			/*
 			 * If multiple threads race to fork, we must ensure
--- a/src/rt-app_parse_config.c
+++ b/src/rt-app_parse_config.c
@@ -610,6 +610,34 @@ parse_thread_event_data(char *name, struct json_object *obj,
 		return;
 	}
 
+	if (!strncmp(name, "fork", strlen("fork"))) {
+
+		data->type = rtapp_fork;
+
+		if (!json_object_is_type(obj, json_type_string))
+			goto unknown_event;
+
+		ref = json_object_get_string(obj);
+
+		i = get_resource_index(ref, rtapp_fork, opts);
+
+		data->res = i;
+
+		rdata = &(opts->resources[data->res]);
+
+		rdata->res.fork.ref = strdup(ref);
+		rdata->res.fork.tdata = NULL;
+		rdata->res.fork.nforks = 0;
+
+		if (!rdata->res.fork.ref) {
+			log_error("Failed to duplicate ref");
+			exit(EXIT_FAILURE);
+		}
+
+		log_info(PIN2 "type %d target %s [%d]", data->type, rdata->name, rdata->index);
+		return;
+	}
+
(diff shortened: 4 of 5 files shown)


qais-yousef:
Yeah I think we certainly have room for improvements but since I wasn't sure which would be the best direction at that time I thought better to push as is and see how can we do better.

I discussed this with Chris and he had a similar suggestion of reusing task. I think we can introduce a new 'subtasks' tag, so we can have something like the following as a description:

"tasks": {
    "task0": { 
         "phases": {
            "p00001": {
                 "fork": "subtask0",
                 .
                 .                                         // Other events inbetween
                 .
                 "fork": "subtask0",
                 "fork": "subtask1",
                 .
                 .
            },
        },
    },
    "task1": { .... },
     .
     .
},
"subtasks": {
     "subtask1": { .... },
     "subtask2": { .... },
     .
     .
}

If we care too much, we could allow

"subtasks": {
     "subtask1": "task0",                    // copy task0 definition
     "subtask2": { .... },
     .
     .
}

which would allow duplicating repetitive tasks. Not sure how difficult it is, so might pursue as a separate improvement as the normal "tasks" definition could benefit from that too .

Thanks for the review.

vingu-linaro:
I agree with Juri. we should not put any constraint with for event. Juri's proposal looks fine with fork event using a parameter to describe with thread should be forked.

jlelli:
> I think we can introduce a new 'subtasks' tag

Mmm, maybe just make it possible to use any task name as a fork candidate.
I still feel that using "instance : 0" property to indicate that a task has not to be created by default should give us all the flexibility that we need. But might be more tricky to implement? Not sure. :-)

qais-yousef:
> I still feel that using "instance : 0" property to indicate that a task has not to be created by default should give us all the flexibility that we need. But might be more tricky to implement? Not sure. :-)

I don't have a strong opinion about it. "instance :0" is good as well although it implies replication rather than spawning e.g: "I want N instances of this tasks". subtasks feels more human readable to me as a worker thread spawned/forked by a main thread. I think code wise it might be clearer to implement too. But again no strong inclination as long as we can achieve the functionality :)

qais-yousef:
OK. Let me see what I can come up with.

But with this approach we can have "instance": 4 which would mean create 4 instances of this task in the beginning.

Thanks for having a look.

jlelli:
Yep. That's how it works today. :-)

qais-yousef:
Huh. I thought you're proposing a new property. Catch the new comer I guess :-D

qais-yousef:
I finally got around doing it today. I did a force push - which as far as I gathered the way to update the pull request.

Lightly tested so feel free to try it out too. I tried sort of a fork bomb and using tasks that has instance: 0. I need to play with it more, but in the time being any feedback is welcome :)

jlelli:
Nice. Thanks for taking time to update this!
I played a little with it. Simple examples seem to work fine.
I've got segfault with something like the following. Could you please test at your end as well?
```
{
	"tasks" : {
		"thread0" : {
			"instance" : 2,
			"loop" : 2,
			"phases" : {
				"light" : {
					"loop" : 10,
					"run" :   3000,
					"timer" : { "ref" : "unique", "period" : 30000 }
				},
				"forks" : {
					"loop" : 5,
					"fork" : "thread1"
				},
				"heavy" : {
					"loop" : 10,
					"run" :   15000,
					"timer" : { "ref" : "unique", "period" : 30000 }
				}
			}
		},
		"thread1" : {
			"instance" : 0,
			"loop" : 1,
			"phases" : {
				"light" : {
					"loop" : 10,
					"run" :   5000,
					"timer" : { "ref" : "unique", "period" : 50000 }
				}
			}
		},
		"thread2" : {
			"instance" : 2,
			"loop" : 2,
			"phases" : {
				"light" : {
					"loop" : 10,
					"run" :   3000,
					"timer" : { "ref" : "unique", "period" : 30000 }
				},
				"forks" : {
					"loop" : 5,
					"fork" : "thread1"
				},
				"heavy" : {
					"loop" : 10,
					"run" :   15000,
					"timer" : { "ref" : "unique", "period" : 30000 }
				}
			}
		}
	},
	"global" : {
		"calibration" : 17,
		"default_policy" : "SCHED_OTHER",
		"pi_enabled" : false,
		"lock_pages" : false,
		"logdir" : "/tmp/",
		"log_basename" : "rt-app1",
		"ftrace" : true,
		"gnuplot" : false
	}
}
```

qais-yousef:
Thanks for giving it a go.

Does the patch below help? I could reproduce but not always. I think the race condition in forking is causing the pthreadd_join() or something related to logging unhappy. If I extend the critical section I think the problem disappeared for me but I'm not sure since it wasn't always to start with..

```
index eb3486a..6b777a1 100644
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -443,8 +443,6 @@ static int run_event(event_data_t *event, int dry_run,
                        pthread_mutex_lock(&fork_mutex);
                        int new_thread_index = nthreads++;
                        threads = reallocarray(threads, nthreads, sizeof(*threads));
-                       pthread_mutex_unlock(&fork_mutex);
-
                        thread_data_t *tdata = malloc(sizeof(thread_data_t));
 
                        if (!threads || !tdata) {
@@ -494,6 +492,8 @@ static int run_event(event_data_t *event, int dry_run,
                        }
 
                        running_threads = nthreads;
+
+                       pthread_mutex_unlock(&fork_mutex);
                }
                break;
        default:
```

jlelli:
This seems to be fixing the issue I was seeing.
I guess we should be unlocking before exiting on errors though (also guess your was a quick try to address the issue so np!).

src/rt-app.c
```diff
@@ -413,6 +432,70 @@ static int run_event(event_data_t *event, int dry_run,
...
+		{
+			log_debug("fork %d", event->count);
+
+			/*
+			 * If multiple threads race to fork, we must ensure
+			 * each one sees a unique index. Hence the lock.
+			 */
+			pthread_mutex_lock(&fork_mutex);
+			int new_thread_index = nthreads++;
+			threads = reallocarray(threads, nthreads, sizeof(*threads));
```
jlelli: Do you think we can pre-calculate the total size of the array (including instance 0 tasks) from the beginning?
It should be more efficient at runtime no?

src/rt-app_parse_config.c
```diff
@@ -890,19 +923,41 @@ parse_tasks(struct json_object *tasks, rtapp_options_t *opts)
...
 	foreach(tasks, entry, key, val, idx) {
 		instance = get_int_value_from(val, "instance", TRUE, 1);
 		opts->nthreads += instance;
+		if (!instance)
+			opts->nzthreads++;
 	}
 
 	log_info(PFX "Found %d tasks", opts->nthreads);
-	opts->threads_data = malloc(sizeof(thread_data_t) * opts->nthreads);
+	log_info(PFX "Found %d ghost-tasks", opts->nzthreads);
```
jlelli: nzthreads as in zero? :-)

qais-yousef:
Yes it was a quick try to verify the fix. I assume it's OK to serialize forking in this way and this won't be considered a limitation. We can probably look at minimizing the critical section duration later if it's become apparent it's a limitation? I can't think why it would be but I tried to keep it small but since we already ran into issue I don't know if it's worth the effort to keep it as small as possible.

src/rt-app.c
qais-yousef: Static allocation was my first approach actually. I had a hard limit on 1024 forks only to minimize allocation (hence uncertainty in execution time). But when I had a chat with Chris about this at some point in the past he had a valid point that when we fork there's an overhead anyway so it could be considered as intrinsic part of the behaviour of this type of loads. Loads that fork have an associated uncertainty of what happens next when the execution of the parent resumes.

I could change it back to static but reallocarray() will not be _that_ slow (don't quote me on that) in the context of the systems rt-app is usually ran on.

qais-yousef: Keep in mind since we can have nested loops (the forked tasks has a phase that forks which in return forks a task that hsa phase that forks etc) - it might get ugly to try to calculate that.

src/rt-app_parse_config.c
qais-yousef: I definitely could have come up with a better name :)

Any suggestions? Maybe nzinstances or nzeroinstances?

src/rt-app.c
jlelli: OK. Not a big deal for the moment. Feel free to keep it as is.

src/rt-app_parse_config.c
jlelli: Name is actually fine with me.
Maybe stick somewhere a comment so that we don't forget what this variable is about?
Guess should be consinsted with how you call them in log_info though (nitpick FTW!).

jlelli:
> I assume it's OK to serialize forking in this way and this won't be considered a limitation. We can probably look at minimizing the critical section duration later if it's become apparent it's a limitation?

Yep, serializing makes sense and I don't think it's a problem. Just make sure we release locks before exiting on error conditions, though.

src/rt-app_parse_config.c
qais-yousef: Does this look better?

`log_info(PFX "Found %d \'\"instance\": 0\' tasks", opts->nzthreads);`

It should be self documented and no need for extra comments I believe.

qais-yousef:
I have pushed the requested changes. I'd still like to do a bit more testing on this. One issue I just realized is that gnuplot for th forked threads. I never used this feature and I think I just circumvented it in the first attempt to get this done. But maybe I should do something about it - or at least document forked threads don'tproduce gnuplots. Maybe they do but I haven't tried this feature of rt-app yet.

Also I found out that we can easily create infinite loop of forks - I have a patch to ignore more forks once we reach a limit. I need think more about this. Any thoughts are of course more than welcome/appreciated.

jlelli:
Thanks!

Regarding gnuplot, I'm not sure it's worth spending time on a feature that I doubt anybody is actually using. @vingu-linaro, are you aware of any users of it?

Regarding infinite loops, I'd vote for exiting with an error, rather than silently ignoring fork bombs, when the max limit is reached. It's easier to debug (if one inadvertently inserted a fork loop in the json description).

qais-yousef:
Cool. Yes exiting on error for infinite loops is a good option. I have got this done and pushed the change.

I am trying to do more testing still. I could see some segfaults sometimes. I'll track it (them) down and if there are no issues after hopefully it should be ready!

qais-yousef:
OK I fixed the segfault and pushed it as a separate commit on top as I think it's worth to be documented in the log.

I think I am reasonably happy with it now. I'll try to see if I can get more people to test it.

Please let me know if there's any more issues to address.

src/rt-app.c
```diff
@@ -37,6 +37,14 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include "rt-app_utils.h"
 #include "rt-app_args.h"
 
+/*
+ * To prevent infinite loops in fork bombs, we will limit the number of
```
derkling: What if if really want to test a fork bomb?

Maybe we can use this just as a default value for a json configuration option... at least we keep the possibilty to set a very high number if that's the intent...

qais-yousef:
Juri, I caught another segfault that is not related to my fork changes. Let me know if you prefer to not lump it with this PR.

qais-yousef:
Is there anything I need to do for this?

src/rt-app.c
```diff
@@ -413,6 +433,83 @@ static int run_event(event_data_t *event, int dry_run,
...
+				exit(EXIT_FAILURE);
+			}
+
+			/*
+			 * If multiple threads race to fork, we must ensure
+			 * each one sees a unique index. Hence the lock.
+			 */
+			pthread_mutex_lock(&fork_mutex);
+			int new_thread_index = nthreads++;
+			threads = reallocarray(threads, nthreads, sizeof(*threads));
```
jlelli: Just discovered (by compiling on Debian 9.5.0) that reallocarray has been introduced in [Glibc 2.26](https://abi-laboratory.pro/?view=changelog&l=glibc&v=2.26). Debian ships with 2.24 and can't find this function.
Guess we should use realloc instead?

qais-yousef: Ah I was looking for realloc() in manpages when I saw it and I didn't realize it's new. Changed it and will update the branch shortly.

Thanks.

doc/examples/tutorial/example9.json
```diff
@@ -0,0 +1,56 @@
...
+	 * Note that thread2 will NOT be instantiated at start time and
+	 * will only run when thread3 forks it. Unlike thread1 which will
+	 * run at the start - since default value of instance property is
+	 * 1 - and will run again when thread3 forks ie: twice in total.
+	 */
+	"tasks" : {
+		"thread1" : {
+			"phases" : {
+				"phase1" : {
+					"run" : 10,
```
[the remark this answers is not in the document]
jlelli: Here and below

doc/tutorial.txt
```diff
@@ -155,7 +160,8 @@ used to create IO-bounded and memory-bounded busy loop. Default value is
 		"gnuplot" : false,
 		"io_device" : "/dev/null"
 		"mem_buffer_size" : 4194304,
-		"cumulative_slack" : false
+		"cumulative_slack" : false,
+		"forks_limit" : 1024
```
jlelli: Not sure what happened, but I could cause a segfault by slightly modifying example9 (notice how I recursively call thread3):
```
{
	/*
	 * A simple example on using fork event.
	 *
	 * thread3 will fork thread1 and thread2 in each of its phases.
	 * Note that thread2 will NOT be instantiated at start time and
	 * will only run when thread3 forks it. Unlike thread1 which will
	 * run at the start - since default value of instance property is
	 * 1 - and will run again when thread3 forks ie: twice in total.
	 */
	"tasks" : {
		"thread1" : {
			"phases" : {
				"phase1" : {
					"run" : 10000,
					"sleep" : 10000
				}
			}
		},
		"thread2" : {
			"instance": 0,
			"phases" : {
				"phase1" : {
					"run" : 20000,
					"sleep" : 20000
				}
			}
		},
		"thread3" : {
			"phases" : {
				"phase1" : {
					"fork": "thread1",
					"run" : 10000,
					"sleep" : 10000
				},
				"phase2" : {
					"fork": "thread3",
					"run" : 20000,
					"sleep" : 20000
				}
			},
			"loop": 10
		}
	},
	"global" : {
		"duration" : 60,
		"calibration" : "CPU0",
		"default_policy" : "SCHED_OTHER",
		"pi_enabled" : false,
		"lock_pages" : false,
		"logdir" : "/tmp/",
		"log_basename" : "rt-app1",
		"ftrace" : true,
		"gnuplot" : false
	}
}
```

doc/examples/tutorial/example9.json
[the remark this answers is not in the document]
qais-yousef: Done.

doc/tutorial.txt
qais-yousef: Hmm I can't reproduce this

qais-yousef: Have you applied/picked up the whole branch? I have fixed a segfault on SIGINT, you haven't hit that by any chance?

jlelli: Mmm, strange. Pretty sure I was on the last complete version, but I'm not able to reproduce it anymore. Fair enough. :-)

doc/tutorial.txt
```diff
@@ -99,7 +99,7 @@ is False.
 * lock_pages : Boolean. Lock the mem page in RAM. Locking the page in RAM
 ensures that your RT thread will not be stalled until a page is moved from swap
 to RAM.  The lock of the page is only possible for non CFS tasks. Default value
-is False.
+is True.
```
jlelli: Forced False for CFS, though. I think we should say that.

src/rt-app.c
```diff
@@ -1250,7 +1250,24 @@ int main(int argc, char* argv[])
...
+		 * complex locking ensure that the original parsed thread_data
+		 * are intact and duplicate them before forking or when we run
+		 * for the first time as in here.
+		 */
+		tdata = malloc(sizeof(thread_data_t));
+		if (!tdata) {
+			perror("Failed to duplicate thread data");
+			exit(EXIT_FAILURE);
+		}
+		memcpy(tdata, &opts.threads_data[i], sizeof(*tdata));
```
jlelli: Don't we need to free this up before thread exit as we do for forked ones?

doc/tutorial.txt
qais-yousef: If there's a monster lurking beneath it'll show its head eventually. But until it does, I'm afraid it'd be hard to do anything :-)

doc/tutorial.txt
qais-yousef: so "Default value is True except for CFS it's FALSE? Does it make sense to change the default to false rather than fixup the documentation?

src/rt-app.c
qais-yousef: Good catch. I think we can free data unconditionally in thread_body() since I changed the logic to always malloc() tdata for forked and unforked tasks.

src/rt-app.c
```diff
@@ -1277,8 +1292,11 @@ int main(int argc, char* argv[])
 		shutdown(SIGTERM);
 	}
 
-	for (i = 0; i < nthreads; i++) {
-		pthread_join(threads[i], NULL);
+	joining = 1;
+	for (i = 0; i < running_threads; i++) {
```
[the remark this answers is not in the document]
jlelli: Mmm, maybe it's just Friday evening :-), but I'm now actually wondering if the pthread_join(s) at the end of main() is ever executed?!

main sleeps for duration and then syncronously calls `shutdown(SIGTERM)` which (if continue_running is still set) wakes everybody up, wait for them and then exits after deconfiguring ftrace (if it was enabled). Shutdown is also executed by async signals.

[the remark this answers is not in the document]
qais-yousef: I ran a lot of tests with no duration was set, and that's how I hit the SIGINT issue. If duration was set, then yeah the pthread_join() logic in main will not be called as shutdown will call exit(EXIT_SUCCESS) once its pthread_join() is done. 

This makes me wonder though if this force wakeup logic called unconditionally makes sense when shutdown() is called from main :-/

doc/tutorial.txt
qais-yousef: I was about to fix this up when I noticed that it does state that "The lock of the page is only possible for non CFS Tasks." I guess is clear enough?

src/rt-app.c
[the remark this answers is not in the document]
jlelli: > with no duration was set

I think this is currently very broken.
So, if no duration was set, main created threads and then waits after joining them.
Then a SIGINT comes, it wakes everybody up and then joins the threads _again_!
According to man pages this is undefined behaviour both if main and shutdown call join concurrently and for whoever comes last. :-/
Maybe we could simply change main to wait for any characted from the user and call shutdown when it receives one (leaving to shutdown the join part)?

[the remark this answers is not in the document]
jlelli: Right, sorry, didn't want to put more work on you. :-)
Maybe the refactoring should still come first though.
Lemme see if I can find some time to do that.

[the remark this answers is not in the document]
qais-yousef: No I am happy to help. But if it's the only thing that blocks this PR to be merged, then I'd appreciate getting it merged first.

I cooked this quick patch though

```
index 661c7d7..cdea703 100644
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -24,6 +24,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 #include <stdlib.h>
 #include <stdio.h>
+#include <stdbool.h>
 #include <string.h>
 #include <fcntl.h>
 #include <math.h>
@@ -545,17 +546,10 @@ int run(int ind,
 	return perf;
 }
 
-static void
-shutdown(int sig)
+static void wakeup_all_threads(void)
 {
 	int i;
 
-	if(!continue_running)
-		return;
-
-	/* notify threads, join them, then exit */
-	continue_running = 0;
-
 	/* Force wake up of all waiting threads */
 	for (i = 0; i <  opts.nresources; i++) {
 		if (opts.resources[i].type == rtapp_wait) {
@@ -565,17 +559,30 @@ shutdown(int sig)
 			pthread_cond_broadcast(&opts.resources[i].res.barrier.c_obj);
 		}
 	}
+}
+
+static void __shutdown(bool force_wake_threads)
+{
+	int i;
+
+	if(!continue_running)
+		return;
+
+	/* notify threads, join them, then exit */
+	continue_running = 0;
+
+	if (force_wake_threads)
+		wakeup_all_threads();
 
 	/*
-	 * If we try to interrupt pthread_join() at the end of main then we
-	 * might end up joining the same thread twice which is undefined
-	 * behaviour and can lead to segfaults.
+	 * Make sure pthread_join() is done once or we hit undefined behaviour
+	 * of multiple simulataneous calls to pthread_join().
 	 *
-	 * Only join if nowhere else started the join process already.
+	 * This is a problem because __shutdown() could be called
+	 * asynchronously from some signals and from main().
 	 */
 	if (joining)
 		return;
-
 	joining = 1;
 
 	/* wait up all waiting threads */
@@ -595,6 +602,12 @@ shutdown(int sig)
 	exit(EXIT_SUCCESS);
 }
 
+static void
+shutdown(int sig)
+{
+	__shutdown(true);
+}
+
 static void create_cpuset_str(cpuset_data_t *cpu_data)
 {
 	unsigned int cpu_count = CPU_COUNT_S(cpu_data->cpusetsize,
@@ -1289,22 +1302,10 @@ int main(int argc, char* argv[])
 		sleep(opts.duration);
 		if (opts.ftrace)
 			log_ftrace(ft_data.marker_fd, "main shutdown\n");
-		shutdown(SIGTERM);
-	}
-
-	joining = 1;
-	for (i = 0; i < running_threads; i++) {
-		int ret = pthread_join(threads[i], NULL);
-		if (ret)
-			perror("pthread_join() failed");
+		__shutdown(false);
 	}
 
-	if (opts.ftrace) {
-		log_ftrace(ft_data.marker_fd, "main ends\n");
-		close(ft_data.marker_fd);
-	}
-	exit(EXIT_SUCCESS);
-
+	__shutdown(false);
 
 exit_err:
 	exit(EXIT_FAILURE);
```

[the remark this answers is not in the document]
qais-yousef: I pushed this as a new patch "Refactor shutdown()"

jlelli:
Thanks for the refactoring patch!
I'd move it to the top of the stack though, it seems a requirement for stabler grounds to me. You could actually also make if a standalone PR if you prefer.

qais-yousef:
Cool. I separated the other fixes into this PR

https://github.com/scheduler-tools/rt-app/pull/66

Once it's merged I'll rebase this series on top

qais-yousef:
I pushed a rebased version on top of latest master after PR 66 was merged.

src/rt-app.c
```diff
@@ -611,7 +696,7 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
 
 	if (!CPU_EQUAL(actual_cpu_data->cpuset, data->curr_cpu_data->cpuset))
 	{
-		log_debug("[%d] setting cpu affinity to CPU(s) %s", data->ind,
+		log_debug("[%d] %s setting cpu affinity to CPU(s) %s", data->ind, data->name,
```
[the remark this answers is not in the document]
qais-yousef: OK. I didn't think this one change is worth a separate patch but will do.

src/rt-app_parse_config.c
```diff
@@ -505,7 +505,7 @@ parse_thread_event_data(char *name, struct json_object *obj,
 		return;
 	}
 
-    if (!strncmp(name, "barrier", strlen("barrier"))) {
+	if (!strncmp(name, "barrier", strlen("barrier"))) {
```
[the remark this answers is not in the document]
qais-yousef: OK I'll split this and the other one into their own white space clean-up patch.

src/rt-app_parse_config.c
```diff
@@ -800,7 +830,7 @@ parse_thread_phase_data(struct json_object *obj,
 }
 
 static void
-parse_thread_data(char *name, struct json_object *obj, int index,
+parse_thread_data(const char *name, struct json_object *obj, int index,
```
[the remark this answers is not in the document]
qais-yousef: Good catch. I think that's a leftover artifact after modifying the way the code works. I'll remove it, and if it's really necessary I'll make sure it is lumped to the correct patch or completely separated.

vingu-linaro:
also patch2 should be the 1st one to make current rt-app support intance 0 before starting to use this new task state

qais-yousef:
Shouldn't it be the other way around? Supporting 0 instance is independent of fork-event. One could use it to temporary 'silence' a task while testing. For example in a test.json one could have 3 tasks all with instance 1, but temporarily decide to switch to testing 2 tasks; so instead of removing the whole task definition from the file one could just set instance to 0, collect the result and then go back to testing the 3 tasks.

IMHO it's a new generic feature that fork takes advantage of but it is not necessarily tied to fork - so it coming first makes more sense as a foundation on which we build the fork event support on.

I've put the white space cleanup and debug statement improvement to be the first patches now.

vingu-linaro:
So I mess up.
 I thought that we were not supporting instance: 0 whereas we supporting it already :-)
patch2: "Parse tasks with 0 instance value" is there so fork event can use task with instance:0 whereas there were not created before, isn't it ?

qais-yousef:
Is there any further work needed for this?

src/rt-app.c
```diff
@@ -415,6 +434,72 @@ static int run_event(event_data_t *event, int dry_run,
...
+	case rtapp_fork:
+		{
+			log_debug("fork");
+
+			/*
+			 * If multiple threads race to fork, we must ensure
+			 * each one sees a unique index. Hence the lock.
+			 */
+			pthread_mutex_lock(&fork_mutex);
+			int new_thread_index = nthreads++;
```
jlelli: So main could already have called join on "old" nthreads before fork events happened, no?
In case duration is not set I mean.

qais-yousef: What's the scenario you're thinking of? I am not sure I understand the problem you're pointing out.

In general, any tasks created by main, pthread_join() will not return until that task has terminated.

For example if we spawn 3 tasks 0, 1 and 2 and task 2 will fork. pthread_join() in main will wait first for 0 to terminate, then 1 then 2. And 2 will not terminate until it has forked all possible tasks. After that pthread_join() will start cycling through all forked tasks with a call to pthread_join().

Assuming we have nested forks; main() will not call pthread_join() on any child thread before its parent has returned/terminated first.

Makes sense? Does this answer your concern somehow?

jlelli: > After that pthread_join() will start cycling through all forked tasks with a call to pthread_join().

Because running_threads is volatile.

Might deserve a comment above the loop in __shutdown() I guess (also because the comment there is at the moment is kind of broken :-).

qais-yousef: It would help if you're more explicit on how the race is happening and what are you requesting to be changed. Also more info on how the comment is broken and what needs to be done will help me to address it appropriately so that we can finally close this long standing poll request.

jlelli: Argh, sorry I wasn't clear.
I meant that you are right in that we shouldn't have any problem (because running_threads is volatile, I added). Just thought that we might want to enrich the comment above [the loop](https://github.com/scheduler-tools/rt-app/blob/master/src/rt-app.c#L490) using runnning_threads, but no big deal if we don't. :-)

jlelli:
LGTM now. @vingu-linaro ?

src/rt-app.c
qais-yousef: Of course. I am more than happy to update it with more info now I understand what you meant :-)

I'll wait to ensure that there are no further comments and will update the PR.

Thanks!

qais-yousef:
```
index 66b7c31..097d206 100644
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -590,7 +590,16 @@ static void __shutdown(bool force_terminate)
        if (pthread_mutex_trylock(&joining_mutex))
                return;

-       /* wait up all waiting threads */
+       /*
+        * Wait for all threads to terminate.
+        *
+        * pthread_join() will block until the process has terminated. If any
+        * thread forks any processes then it'll update running_threads before
+        * it returns and since running_threads is volatile the for() loop is
+        * guaranteed to check against the updated value in each iteration.
+        * Hence we are guaranteed to wait for all forked threads beside the
+        * originally created ones at startup time.
+        */
        for (i = 0; i < running_threads; i++)
        {
                int ret = pthread_join(threads[i], NULL);
```

I'll fold this up to the "Add a new 'fork' event" commit.

jlelli: Nice, thanks!

qais-yousef:
I just found out an interesting property in json definition regarding duplicate keys

> Also, as Serguei pointed out in the comments: ECMA-262 "ECMAScript® Language Specification", reads:

    In the case where there are duplicate name Strings within an object, lexically preceding values for the same key shall be overwritten.

    (in other words last-value-wins).

https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object

So if we have

```
.
.
"phase1": {
    fork: "thread0",
    fork: "thread1"
},
.
.
```

Only the last fork will be registered - json-c implements this behaviour at least :-/

Need to find a way to deal with this!

vingu-linaro:
Yes this is a know behavior of json.
That's why we have the Workgen script that ensure uniqness of json node

qais-yousef:
OK. I wasn't aware of it to be honest. One way to deal with this is to put each fork in its own phase. Or allow the property to be 'wild matched' e.g:

```
"phase_fork": {
    fork_1: "thread1",
    fork_2: "thread2",
    forkWHATEVER: "thread3"
}
```

vingu-linaro:
Yes you don't have to do anything

derkling:
On 15-Oct 09:09, Qais Yousef wrote:
> I just found out an interesting property in json definition regarding duplicate keys
> 
> > Also, as Serguei pointed out in the comments: ECMA-262 "ECMAScript® Language Specification", reads:
> 
>     In the case where there are duplicate name Strings within an object, lexically preceding values for the same key shall be overwritten.
> 
>     (in other words last-value-wins).
> 
> https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object
> 
> So if we have
> 
> ```
> .
> .
> "phase1": {
>     fork: "thread0",
>     fork: "thread1"

Not really sure... but... the simplest solution comes to mind is:

    "phase1": {
        fork: "thread0,thread1"

Does that makes any sense ?

-- 
#include <best/regards.h>

Patrick Bellasi

qais-yousef:
Yes this sounds better actually. Json accepts arrays to describe multi value key.

But if this PR is OK as is I'd rather merge it already and push any improvements on top separately.

vingu-linaro:
We have to support :
"phase1": {
     fork: "thread0",
    "run": 1000,
     fork: "thread1"
}

So we can easily support 
"phase1": {
     fork: "thread0",
     fork: "thread1"
}

Whereas  supporting  fork: "thread0, "thread1"
or probably better fork: ["thread0", "thread1"]

means to check the type of the fork node.

And for now I'm not sure to see the interest compare to the added complexity

jlelli:
> And for now I'm not sure to see the interest compare to the added complexity

Right. I'd leave this for another time. Maybe create an issue so that we don't forget?

qais-yousef:
@vingu-linaro I addressed your requested comments. Are you waiting to get through the code again for a second review or you're happy with it now?

doc/tutorial.txt
```diff
@@ -489,6 +489,10 @@ taskB ...-------------------|resume taskA|run 10    |--------------------
 * yield: String. Calls pthread_yield(), freeing the CPU for other tasks. This has a
 special meaning for SCHED_DEADLINE tasks. String can be empty.
 
+* fork: String. Instead of creating all tasks at start time, you can fork one
+at any point of time using this event. The name of the forked task must match
+one of the defined tasks.
+
```
vingu-linaro: you have to update the description of "instance". Add that 0 will now parse and create data structure of the task but will not create any pthread.

src/rt-app.c
```diff
@@ -415,6 +434,72 @@ static int run_event(event_data_t *event, int dry_run,
 			pthread_yield();
 		}
 		break;
+	case rtapp_fork:
+		{
+			log_debug("fork");
```
vingu-linaro: log of other events provide more information: log_debug("fork %s", rdata->res.fork.ref)

src/rt-app.c
```diff
@@ -415,6 +434,72 @@ static int run_event(event_data_t *event, int dry_run,
...
+
+			if (!threads || !tdata) {
+				log_error("Failed to allocate memory for a new fork: %s", rdata->res.fork.tdata->name);
+				pthread_mutex_unlock(&fork_mutex);
+				exit(EXIT_FAILURE);
+			}
+
+			/* Find the thread data associated with this fork */
+			if (!rdata->res.fork.tdata) {
+				rdata->res.fork.tdata = find_thread_data(rdata->res.fork.ref, &opts);
```
vingu-linaro: worth to add a comment to mention that you keep a pointer to the tdata to prevent re doing it for next loop.
Because apart for this optim, the pointer is not needed.
This raise a question ? why not setting this pointer when parsing the json file instead of waiting for the 1st event ?

src/rt-app.c
```diff
@@ -415,6 +434,72 @@ static int run_event(event_data_t *event, int dry_run,
...
+			/* Mark this thread as forked */
+			tdata->forked = 1;
+			/* update the index value */
+			tdata->ind = new_thread_index;
+
+			/* Make sure each forked thread has a unique name */
+			int string_size = strlen(tdata->name) + 1 /* NULL */ + 10 /* postfix */;
+			char *unique_name = malloc(string_size);
+			if (unique_name) {
+				snprintf(unique_name, string_size, "%s%03d", tdata->name, rdata->res.fork.nforks++);
```
vingu-linaro: this is different from default naming convention.
I mean if we have "instance" :2, the name of the tasks will be the same for the 2 tasks
but if we use "instance" :0 and then call twice "fork" : "taskname", the name of the 2 tasks will be different. We have to be consistent with both case. At now, we use data->ind which is unique to make difference between the task.
Futhermore, if I describe 11 tasks in my json file form "task1" to "task11" and then I fork "task1", i will have 2 tasks with same name "tasks11" ... So you are not unique 
I understand the interest of uniqueness but for now data->ind ensure it. I fyou want unique name, you needs more complex mecanism and you will have to extend it to tasks create at beg of the use case and not only for forked task

doc/tutorial.txt
qais-yousef: Added that to the instance parameter description and folded it into the (Parse tasks with 0 instance value) commit

src/rt-app.c
qais-yousef: Done

doc/tutorial.txt
```diff
@@ -489,6 +489,10 @@ taskB ...-------------------|resume taskA|run 10    |--------------------
 * yield: String. Calls pthread_yield(), freeing the CPU for other tasks. This has a
 special meaning for SCHED_DEADLINE tasks. String can be empty.
 
+* fork: String. Instead of creating all tasks at start time, you can fork one
```
vingu-linaro: I thought i made a comment on "instance":0 but can't find it anymore. So this comment maybe duplicated with another one:
You should update the description of "instance" for saying that "instance": 0 will now create the data of the task but will not create the thread. This is a change compare to current behavior. at now "instance":0 can be used to remove a task for the use case

src/rt-app.c
qais-yousef: Because we have to wait for all threads to be parsed before we can search for the reference. We could do something at parse time probably but this seems a lot more efficient and simpler to me.

I'll add comments.

src/rt-app.c
qais-yousef: I agree about the inconsistencies in behavior if we use "instance": 2.

> Futhermore, if I describe 11 tasks in my json file form "task1" to "task11" and then I fork "task1", i will have 2 tasks with same name "tasks11

This is not true as we use "%s03d" so we will have task1001 not task11. But yes we are not guaranteed uniqueness if something called a task1001.

So the proposal is to always append data->ind to the task name? The advantage of nforks used now is that it'll increment each fork independently.

for example if you have 2 fork events in a task; `"fork": "thread1"` and `"fork": "thread2"` each fork will start appending a 'unique' number starting from 000. This will be very helpful when looking at traces and trying to identify a particular run. Using ind there will be a bit of homework to be done to find out what's the name of the 8th fork of a particular event for example.

Maybe we can have a pattern like this: `"%s-%d-%04d", name, ind, nforks`?

Any thing forked at the beginning will only have `"%s-%d", name, ind` only

doc/tutorial.txt
qais-yousef: yeah same here I received the emails but took me a while before I could see the actual comments on the page - github glitch probably..

qais-yousef:
I updated the branch with the fixups on top to facilitate reviewing - once they are ok I'll squash them and push the final branch for merging

src/rt-app.c
vingu-linaro: yes but we can handle that. And I think that we have to.
With the new mechanism of duplicating thread_data, we have a waste a memory and structure. 
In current version, we create a thread_data per task and use this structure directly in the thread.
With fork feature, we are duplicating the thread_structure and use the copy in the thread but we are still creating one thread_structure per instance. This means with "instance" : 1000 as an example, we will allocate a thread_data array of 1000 cells, parse 1000 times the json file to fill the structures but then duplicate them so we have 2000 structures and half of them are unused.
The right way to do would be to create only 1 thread_data structure per task node and then duplicate it when we create thread.
In addition, we can also split the parsing in 2 steps:
1st step parse all task nodes and create this main structures
2nd step, parse the phases and fill the array of phases and events. At that time, you can directly set the pointer of fork event instead of waiting for the 1st event to run

qais-yousef: >  parse 1000 times the json file to

This will never happen, we parse the json file once, no? I just extended parse_tasks() but didn't change the json file is parsed.

We can optimize the parsing loop to ignore instances; I'll try to spend a little time on that. But FWIW the sizeof(thread_data_t) is 160 bytes. If we spawn 4000 (which is very unlikely for the purpose of rt-app other than to create a forkbomb maybe) we will need ~640KiB. If this comment appeared earlier it would have been easier to address - but given the long time since I last touched it I agree this is suboptimal but for all purposes of rt-app this is just fine and will hardly be a limitation except for the tiniest of systems where Linux probably can't run on them anyway.

vingu-linaro: no look at parse_tasks, we parse the task node for each instance to create the thread_data of each thread
For fork, you change the behavior and you just need to create one instance because you duplicate it for each thread. Do it completely and cleanup unused data.
I'm sorry about the delay in review but can't accept sub optimal behavior like that even if it would have been only 1 Byte per instance. 
You change the init sequence for the fork event but you have to do that completely and cleanly and not changing part of it and let dead code or data which means:
- remove useless creation of structure
- don't resolve at runtime something that could have been done during the parse of the json file. all other events setup everything during the parse of the json file so we need to be consistent. Furthermore, you create a difference between the 1st and the next events without any good reason apart from reducing the code refactoring

qais-yousef: No one wants to waste anything. But when the review process is very slow and disjointed - it seems more fruitful for everyone's time to accept what we have and push improvements in top so that progress is made. Especially the issue highlighted here is a no-issue from practical point of view and meets the requirement of rt-app adequately.

So once I fix this will this be merged? A recurring theme in the review process is to pop up to request some changes then disappear - after that appear again to request changes in different area and this is wasting my time as I want to address all comments in one go. If you want to look at the code again please do so and let me know so that I can address everything in one go. 

I appreciate all comments given and time taken to look at this request by the way - I don't want the above to sound ungrateful in anyway :)

src/rt-app.c
```diff
@@ -611,8 +748,8 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
 
 	if (!CPU_EQUAL(actual_cpu_data->cpuset, data->curr_cpu_data->cpuset))
 	{
-		log_debug("[%d] setting cpu affinity to CPU(s) %s", data->ind,
-			actual_cpu_data->cpuset_str);
+		log_debug("[%d] %s setting cpu affinity to CPU(s) %s",
+			data->ind, data->name, actual_cpu_data->cpuset_str);
```
vingu-linaro: This is not linked to fork event. Please keep it separate

qais-yousef: This was addressed already

https://github.com/scheduler-tools/rt-app/pull/53/commits/7fc5dc6e52c32f7abc9190ebc4e315fcb2ad74e3

src/rt-app.c
qais-yousef: This is addressed here now:

https://github.com/scheduler-tools/rt-app/pull/53/commits/07a1138edf79c7cfe17ed347b0a198ecf7be194e

qais-yousef:
I've updated the branch which addresses all review comments.

I kept the changes as separate comments on top as fixups which will be squashed once the review is done.

The last commit to optimize memory usage I rather keep as a single commit rather than squashing it into the fork: event one for 2 reasons:

1. It is a documentation of how the code evolved. There's valuable information for new comers when they see this as a separate commit.

2. It would be easier to revert it without reverting all of fork support if any regression was caused by this change.

If there are no further comments on my new commits please let me know so I can do the squashing and update the branch.

qais-yousef:
Ping. I addressed all comments, hopefully it's good enough to be merged now? Keep in mind I still need to squash some fixups before merging, I just left them as is to facilitate incremental review.

src/rt-app.c
qais-yousef: I don't understand what you want. Can you clarify please?

This was already split into its own commit

https://github.com/scheduler-tools/rt-app/pull/53/commits/7fc5dc6e52c32f7abc9190ebc4e315fcb2ad74e3

Are you looking at the latest version of the patchsets? I have to say github view of the incremental changes confuses me.

qais-yousef: Do you want the commit to be dropped altogether, is this is what you're saying?

src/rt-app.c
```diff
@@ -1168,14 +1306,49 @@ int main(int argc, char* argv[])
...
+			/* And to update the name with something unique */
+			thread_data_set_unique_name(tdata, 0);
+
+			setup_thread_logging(tdata);
+
+			if (pthread_create(&threads[ind],
+					  NULL,
+					  thread_body,
+					  (void*) tdata))
+				goto exit_err;
```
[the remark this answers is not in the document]
qais-yousef: OK shall do.

src/rt-app_parse_config.c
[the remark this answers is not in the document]
qais-yousef: This was split into its own commit too.

https://github.com/scheduler-tools/rt-app/pull/53/commits/76102ecddbafde72fa9a72c14025b2f270c6e16f

Can you clarify what 'remove' means here too please?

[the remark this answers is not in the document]
vingu-linaro: We try to prevent mixing patches from different subject in the same patchset/pull in case some revert are needed as an example. These patches should be in a dedicated pull request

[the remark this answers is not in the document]
qais-yousef: I'm sorry that doesn't make sense to me as they are very tiny improvements. I'll drop them anyway. But I am afraid I won't be sending a new PR just for them.

qais-yousef:
Hi Vincent. I dropped the two cleanup  patches and added a fixup commit at the top to address your last request of wrapping the thread creation in a function.

Lightly tested only, but if you're OK with it I'll give it additional testing before the end of the week and  squash the fixup commit - given you're happy with the current state of the code and have no additional comments.

Thanks.

src/rt-app.c
```diff
@@ -108,6 +108,64 @@ static void thread_data_set_unique_name(thread_data_t *tdata, int nforks)
...
+
+	if (!td) {
+		log_error("Failed to fork new thread, passed NULL thread_data_t: %s", td->name);
+		return -1;
+	}
+
+	tdata = malloc(sizeof(thread_data_t));
+	if (!tdata) {
+		log_error("Failed to allocate memory for a new fork: %s", td->name);
+		return -1;
```
vingu-linaro: It's not only for fork event so I prefer to keep the generic log : "Failed to duplicate thread data: %s", td->name

That's the only comment i have. Otherwise looks good to me

qais-yousef: Indeed. Copy paste error from my side. Fixed it locally and will hammer it today and tomorrow and push the final branch for merging on Friday.

Thanks a lot for you time!

qais-yousef:
I rebased and pushed with all fixes applied. I tried it on x86 and arm and didn't notice any problem. So should be good to merge now.

Thanks

vingu-linaro:
The result of the patchset looks good to me but there is problem with patches:
- 1st patch doesn't compile: rt-app.c:66:39: error: ‘rtapp_options_t {aka struct _rtapp_options_t}’ has no member named ‘nzthreads’
- commit of 1st patch mentioned:  "Use "instance": 0 property to prohibit a task from being created at
    start time. You can still refer to this task to fork it at runtime." But this is added by patch 3

If you don't want to bother with creating separate patches, you can probably squash patch 1,3 and 5

qais-yousef:
I fixed the compilation error and removed that paragraph from the commit message. I checked that each individual patch compiles correctly too - I hate to break bisection as well. Good catch.

qais-yousef:
Thanks for all reviewers. I think this can be merged in now if one of the maintainers would like to hit the merge button.

jlelli:
Awesome! Thanks a lot for this. And thanks to the reviewers as well. :-)

PR #56, opened 2018-09-07
doc: fix build instructions

v0lker:


doc/tutorial.txt
```diff
@@ -66,7 +66,7 @@ $ tree -d
 
 the configure step becomes
 
-./configure --host=aarch64-linux-gnu LDFLAGS="--static -L./../json-c/." CFLAGS="-I./.." --with-deadline
+./configure --host=aarch64-linux-gnu LDFLAGS="-static -L"`pwd`"/../json-c" CFLAGS="-I./../json-c" --with-deadline
```
vingu-linaro: Please use $PWD instead of "`pwd`" it's easier to read and understand

jlelli: Mmm, it's working here only if I keep `--static` and `-L"`pwd`"/json-c"`, basically

```
./configure --host=aarch64-linux-gnu LDFLAGS="--static -L"`pwd`"/json-c" CFLAGS="-I./../json-c/" --with-deadline
```

vingu-linaro: there is a mismatch between CFLAGS="-Ipath to parent of json repo" and CFLAGS="-I./../json-c"

vingu-linaro:
Could you also add in the commit message that it's not only the absolute path for LDFLAG but the -static and also you have changed the path.

vingu-linaro:
sorry wrong button

doc/tutorial.txt
v0lker: ok

v0lker: according to gcc docs (https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html) , it's `-static`, not `--static`, so i took a look and came to the conclusion that actually both `-static` and `--static` are wrong. proposed solution at the end, keep reading if you're interested.

diffing the make invocations:
```
-/bin/sh ../libtool  --tag=CC   --mode=link gcc  -I./../json-c  -static [...]
-libtool: link: gcc -I./../json-c -o rt-app [...]
+/bin/sh ../libtool  --tag=CC   --mode=link gcc  -I./../json-c  --static [...]
+libtool: link: gcc -I./../json-c --static -o rt-app [...]
```

notice that in the `-static` case, the flag is missing from the linker invocation and in the `--static` case it is passed through (and gcc doesn't complain)

looking at libtool (https://www.sourceware.org/autobook/autobook/autobook_59.html), it is stripping `-static`, which is a flag that it consumes itself:
```‘-static’
    This switch is similar to ‘-all-static’, except that it applies to only the uninstalled Libtool libraries in the build tree. Where possible the static archive from these libraries is used, but the default linking mode is used for libraries which are already installed. ```

the equivalent of gcc's linker invocation with `-static` seems to be `-all-static`:
```‘-all-static’
    Always choose static libraries where possible, and try to create a completely statically linked executable.```

(when i earlier used `-static`, i ended up with a dynamic executable, which i failed to check.)
also see "why does libtool strip link flags...": https://www.gnu.org/software/libtool/manual/html_node/Stripped-link-flags.html

it seems the properest(TM) way to do this is to do leave the static-ness to libtool (like it is apparently meant to be - libtool doesn't like the user messing with linker flags) and do the following:

```
./configure LDFLAGS="-L$PWD/../json-c" CFLAGS="-I./../json-c" --with-deadline"
AM_LDFLAGS="-all-static" make
```

going to push the fixed version, please let me know your thoughts.

Comment (v0lker) [reply doc/tutorial.txt]: well spotted, thanks.

File: doc/tutorial.txt [hunk 2]
```diff
@@ -66,7 +66,7 @@ $ tree -d
 
 the configure step becomes
 
-./configure --host=aarch64-linux-gnu LDFLAGS="--static -L./../json-c/." CFLAGS="-I./.." --with-deadline
+./configure --host=aarch64-linux-gnu LDFLAGS="-L$PWD/../json-c" CFLAGS="-I./../json-c" --with-deadline
```
jlelli: This looks now good to me and generates a statically linked binary.
However, with a directory tree like in the above example LDFLAGS is not correct, as it should be
```
./configure --host=aarch64-linux-gnu LDFLAGS="-L$PWD/json-c" CFLAGS="-I./../json-c" --with-deadline
```

v0lker: you're right, i have json-c in the parent directory (i don't like nesting unrelated projects' build directories). could either fix the tree or the invocation, whichever you prefer.

just for clarification, i think that should have been
```
./configure --host=aarch64-linux-gnu LDFLAGS="-L$PWD/json-c" CFLAGS="-I./json-c" --with-deadline
```
instead of what you wrote above..(note the CFLAGS)

jlelli: Let's go with yours. Please update the text accordingly.

v0lker: done that. is it ok?

doc/tutorial.txt
```diff
@@ -40,33 +40,31 @@ For rt-app:
...
+    |-- libdl
+    |-- m4
+    `-- src
+
 
 the configure step becomes
 
-./configure --host=aarch64-linux-gnu LDFLAGS="--static -L./../json-c/." CFLAGS="-I./.." --with-deadline
+cd rt-app
+./configure --host=aarch64-linux-gnu LDFLAGS="-L$PWD/../json-c" CFLAGS="-I./../json-c" --with-deadline
```
jlelli: Sorry for getting back again on this, but with a tree like the above this triggers the following
```
rt-app_parse_config.c:28:10: fatal error: json-c/json.h: No such file or directory
 #include <json-c/json.h>
          ^~~~~~~~~~~~~~~
compilation terminated.
```
Changing the configure step to
```
./configure --host=aarch64-linux-gnu LDFLAGS="-L$PWD/../json-c" CFLAGS="-I./../.." --with-deadline
```
Does the trick, so it looks like it's actually
```
./configure --host=aarch64-linux-gnu LDFLAGS="-L<absolute path to json repo>/json-c" CFLAGS="-I<path to parent of json repo>" --with-deadline
```

jlelli:
Yep this works and it seems the best solution for me as well.

doc/tutorial.txt
jlelli: Yep this works and it seems the best solution to me as well.

doc/tutorial.txt
```diff
@@ -40,33 +40,31 @@ For rt-app:
 
 export ac_cv_lib_json_c_json_object_from_file=yes
 ./autogen.sh
-./configure --host=aarch64-linux-gnu LDFLAGS=" --static -L<path to parent of json repo>/json-c/." CFLAGS="-I<path to parent of json repo>" --with-deadline
-make
+./configure --host=aarch64-linux-gnu LDFLAGS="-L<absolute path to json repo>/json-c" CFLAGS="-I<path to json repo>" --with-deadline
```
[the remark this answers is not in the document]
v0lker: you're absolutely right. sorry, my bad.

doc/tutorial.txt
v0lker: i tested it on my machine but i had the headers installed in /usr/include.. sorry.

v0lker:
sorry, this is turning into a longer story than i originally intended. there seems to be a new (actual) issue: since i was using the one supplied by the system package, i didn't realise that the current version of json-c doesn't define 'TRUE' in json_object.h any longer.

would you rather like in `src/rt-app_parse_config.c`:
```
#ifndef TRUE
#include <json_object.h>
#define TRUE ((json_bool)1)
#define FALSE ((json_bool)0)
#endif
```

OR

```
#include <stdbool.h>
```
and rename all `TRUE`s to `true`

it is 2018, so we can assume C99 support and we can assume that JSON truth will be the same as the canonical one -?

OR

should we use `0` and `1` as the json-c people seem to intend ("Remove TRUE and FALSE defines, use 1 and 0 instead" 0992aac61f8b087efd7094e9ac2b84fa9c040fcd)

PR #58, opened 2018-09-07
actually use the loop value for phase-less threads, changes in documentation that hopefully clarify how looping works.

v0lker:
if a thread without a phase is declared, the 'loop' setting was ignored and '1' was used, instead. (i am assuming that this should apply to the thread, is that correct?)


v0lker:
if there is no phase property, there is ambiguity which element the loop property belongs to and i interpreted it the wrong way around.

PR #60, opened 2018-09-11
small documentation / plotting changes

v0lker:
some of the units seemed wrong in the plots. i think the choice of ms in one and us in the other x-axis scale seems a bit odd, that could still be harmonised if anybody cared enough.

some cosmetics for the plots again.

it would be nice if it were the same and easier to parse for humans (no point in having 7 labels, with just 2 different values: 1.19039e9 us and 1.1904e9 us...?)... but maybe that's too much effort for too little gain (who actually uses these plots instead of something like LISA?)

in the documentation, the parameter to the run event used the wrong unit (unless i'm missing something) and i found it hard to understand, so i took a stab at rewording it.

could split this PR up, if you let me know which commits you actually like - ?


doc/tutorial.txt
```diff
@@ -310,11 +310,15 @@ events are simple action that will be performed by the thread or on the
 thread. They have to be listed by execution order.
 
 * run : Integer. Emulate the execution of a load. The duration is defined in
-usec but the run event will effectively run a number of time a loop that waste
-cpu cycles. When the run event is executed, the duration is transformed in a
-number of loop thanks to the ns per loop value (see calibration). This way of
-working enables to emulate a load with a duration that will vary with the
-frequency or the compute capacity of the CPU.
+ns but the run event will effectively run a number of time a loop that waste
```
jlelli: It's actually usec.

doc/examples/template.json
```diff
@@ -7,17 +7,24 @@
...
 		"thread0" : {
-			"instance" : 1,
-			"loop" : -1,
-			"run" :   10000,
-			"sleep" : 0,
+			"instance" : 1,  /* number of threads to be created */
+			"loop" : -1,  /* loop forever (see 'duration' below) */
+
+			"run" :   10000,  /* when woken up,
+				execute 10000 / pLoad loops (note: rounding) */
```
vingu-linaro: the comment is confusing. /* when woken up, run for 10000 usec. The duration is converted in a number of 10000/pload loops ( floor rounded) */

doc/examples/template.json
```diff
@@ -7,17 +7,24 @@
...
+
+			"sleep" : 0, /* after running, do not sleep, (then
+				go to sleep until timer fires) */
```
vingu-linaro: In fact this event doesn't generate anything because sleep is 0 and is somewhat useless.

doc/examples/template.json
```diff
@@ -7,17 +7,24 @@
...
+
 			"timer" : { "ref" : "unique", "period" : 100000 }
+			/* trigger thread every 100000 us: */
```
vingu-linaro: /* wake up thread every 10000 us */ ?

src/rt-app.c
```diff
@@ -1100,9 +1100,10 @@ int main(int argc, char* argv[])
 				"set key outside right\n"
 				"set title \"Measured %s Loop stats\"\n"
 				"set xlabel \"Loop start time [msec]\"\n"
```
vingu-linaro: ins't the start time in usec as well ?

doc/tutorial.txt
v0lker: is it? i have a calibration report of pLoad=45ns. setting 'run' to 10000 gives me a 'perf' column of 222, so that means 222 runs * 45ns / run = 9990ns, hence i  assumed it is in ns, not us - ?

doc/examples/template.json
v0lker: i thought it's clearer the way i wrote it (because you don't actually run for the exact amount of time) but then i'm biased.

doc/examples/template.json
v0lker: ok, changing that accordingly

doc/examples/template.json
v0lker: ok. 100000?

src/rt-app.c
v0lker: not on that graph, i think.

doc/examples/template.json
vingu-linaro: yes sorry  100000us. one 0 was missing

doc/tutorial.txt
vingu-linaro: The event run is specified in us. The pload uses ns for precision

v0lker: 222 runs of 45ns are not 10000us but 10000ns, though - which seems to suggest that the parameter is in ns. apologies if this is really obvious but... what am i missing?

vingu-linaro: look at the code : 
```
	/*
	 * Run for the remainig exec (if any).
	 */
	load_count = (exec * 1000)/p_load;
	waste_cpu_cycles(load_count);

```
exec is the duration in usec which is then multiply by 1000 to becomes ns before being divided by p_load

but we return exec / p_load number of loop instead of (exec * 1000)/p_load

I can't remember exactly but don't we discuss that soem time ago Juri ? I think it was because of some overflow poblems that we return exec / p_load instead of (exec * 1000)/p_load

v0lker: ahh. i should've paid more attention. i just looked at the beginning, where it said
```
194         perf = exec / p_load; 
```

 thanks! so, perf in the log file is actually in 1000 pLoad units. correct? i think that needs adding to the documentation...

vingu-linaro: yes. I agree

v0lker: updated the doc again and hopefully got the units right, this time...

doc/examples/template.json
```diff
@@ -7,17 +7,23 @@
...
+		/* no 'loop' value at task level means loop indefinitely (-1) */
 		"thread0" : {
-			"instance" : 1,
-			"loop" : -1,
-			"run" :   10000,
-			"sleep" : 0,
+			"instance" : 1,  /* number of threads to be created */
+			"loop" : -1,  /* loop forever (see 'duration' below) */
+
+			"run" :   10000,  /* create load [ns]: when awaken,
```
jlelli: run it's in [us] (as stated in tutorial.txt)

v0lker: sorry, missed that one.

Issue #59, opened 2018-09-07
change log level per rt-app invocation


v0lker:
currently, it is a compile-time option which i find a bit tedious. could have a command line flag (how about -l ?). i'm assuming that the overhead is acceptable. could provide a PR, if this feature isn't disliked - ?

jlelli:
It might be handy. Maybe just follow the -v -vv -vvv common pattern?
Also maybe keep a config option to completely disable logging? Just in case overhead is found not to be so negligible.

v0lker:
-v is already taken for 'version', so i'd rather leave it in peace. any other suggestion?

as for the overhead and completely disabling:
going from setting the debug level at compile time to doing it dynamically, we'd incur the overhead of an integer comparison for every log_*() invocation. the branch predictor should be very effective here (i think) as the result is always the same.

right now i can think of only one way to have even less overhead and still log dynamically: rewriting the log function in the process' .text in memory on start-up (replacing the function calls with nops). i don't think it is worth the effort, though.

what would you suggest?

jlelli:
OK. Please go for `-l --log` and don't worry about the overhead. We will figure it out later if it turns out to be a real problem, I guess.

v0lker:
remaining task is to allow a string instead of the numerical value for the `-l` option. i'll provide a PR soon, unless someone else is eager...

v0lker:
https://github.com/scheduler-tools/rt-app/pull/68 implements the remaining issue

PR #61, opened 2018-09-11
add option to change debug/log level per invocation

v0lker:
proposed implementation of https://github.com/scheduler-tools/rt-app/issues/59

### DIFF
--- a/src/rt-app_args.c
+++ b/src/rt-app_args.c
@@ -26,14 +26,15 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include <getopt.h>
 
 #include "rt-app_parse_config.h"
+#include "rt-app_utils.h"
 
 char help_usage[] = \
-"Usage: rt-app <taskset.json>\n"
+"Usage: rt-app [-l <debug_level>] <taskset.json>\n"
 "Try 'rt-app --help' for more information.\n";
 
 char help_full[] = \
 "Usage:\n"
-"      rt-app <taskset.json>\n"
+"      rt-app [-l <debug_level>] <taskset.json>\n"
 "      cat taskset.json | rt-app -\n\n"
 "taskset.json is a json file describing the workload that will be generated"
 "by rt-app.\n\n"
@@ -41,8 +42,10 @@ char help_full[] = \
 "In the second example, rt-app reads the workload description in json format\n"
 "through the standard input.\n\n"
 "Miscellaneous:\n"
-"  -v, --version             display version information and exit\n"
-"  -h, --help                display this help text and exit\n";
+"  -v, --version      display version information and exit\n"
+"  -l, --log          set verbosity level (10: ERROR/CRITICAL, 50: NOTICE (default)\n"
+"                                    75: INFO, 100: DEBUG)\n"
+"  -h, --help         display this help text and exit\n";
 
 void
 usage(const char* msg, int ex_code)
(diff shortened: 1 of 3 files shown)


src/rt-app_args.c
```diff
@@ -66,12 +69,10 @@ parse_command_line(int argc, char **argv, rtapp_options_t *opts)
 	struct stat config_file_stat;
 	int c;
 
-	while (1) {
-		c = getopt_long(argc, argv, "hv", long_args, 0);
-		if (c == -1)
-			break;
-
+	while (-1 != (c = getopt_long(argc, argv, "hvl:", long_args, 0))) {
```
jlelli: Why changing also the way the loop is written?

src/rt-app_args.c
```diff
@@ -42,6 +43,7 @@ char help_full[] = \
 "through the standard input.\n\n"
 "Miscellaneous:\n"
 "  -v, --version             display version information and exit\n"
+"  -l, --log                 set log level (0...100 recommended, 50 default)\n"
```
jlelli: Why recommended?

jlelli:
Apart from the nitpick, thanks for this LGTM. @vingu-linaro ?
Bonus point would be to add a commit on top implementing "translation" between magic numbers and more meaningful strings, e.g.. instead of
```
rt-app -l 75
```
would be nice to be able to write
```
rt-app -l info
```

src/rt-app_args.c
v0lker: because other values don't make sense. actually, the lowest value that makes sense is 10 (ERROR and CRITICAL), the highest is 100.

i was trying to keep within the string length, but could elaborate, if you like...

jlelli: Ah OK sorry.. I misinterpreted as if 100 was the recommended value.
Maybe we could just remove "recommended" and check that user don't pick something outside the range?

v0lker:
i agree with the point about meaningful values. that should be done properly (with an array of LOG_LEVELs and their associated strings), though. would it be ok if i just state the levels explicitly in the help text, in the meantime?

jlelli:
> would it be ok if i just state the levels explicitly in the help text, in the meantime?

Sure. Please add INFO and NOTICE in the help text above.

v0lker:
done that...

src/rt-app_args.c
```diff
@@ -67,11 +71,13 @@ parse_command_line(int argc, char **argv, rtapp_options_t *opts)
 	int c;
 
 	while (1) {
-		c = getopt_long(argc, argv, "hv", long_args, 0);
+		c = getopt_long(argc, argv, "hvl:", long_args, 0);
 		if (c == -1)
 			break;
 
 		switch (c) {
+		case -1:
```
jlelli: Isn't this caught by the `if (c == -1)` above?

src/rt-app_args.c
```diff
@@ -83,6 +89,20 @@ parse_command_line(int argc, char **argv, rtapp_options_t *opts)
...
+				char *endptr;
+				long int ll = strtol(optarg, &endptr, 10);
+				if (*endptr) {
+					usage(NULL, EXIT_INV_COMMANDLINE);
+					break;
+				}
+				log_level = atoi(optarg);
+			}
+			break;
+		case '?':
```
jlelli: Why adding this?

src/rt-app_args.c
v0lker: yes, it is...

src/rt-app_args.c
v0lker: you're right, there is no point..

jlelli:
It looks now good to me! Thanks!
@vingu-linaro you also OK for merging?

vingu-linaro:
LGTM too.
I agree that adding meaningful naming for log level in a next step, would be really helpful.

PR #65, opened 2018-09-17
parse_config: Define TRUE and FALSE if not already defined by json-c.

jlelli:
Some newer versions of json-c library stopped defining TRUE and FALSE
and this causes a compile issue.

Fix it my defining those if not yet defined by json-c (to be back and
forward compatible with the library).


jlelli:
This is from @pino-kim , just squashed commits and changed commitlog.
Please have a look.

pino-kim:
@jlelli 
I'm fine. It's look better.

PR #66, opened 2018-09-19
Misc enhancements

qais-yousef:
Few fixes and improvements caught while developing fork-support. Since they grew to a bit they warrant their own PR now

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -24,6 +24,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 #include <stdlib.h>
 #include <stdio.h>
+#include <stdbool.h>
 #include <string.h>
 #include <fcntl.h>
 #include <math.h>
@@ -46,6 +47,7 @@ static int p_load;
 rtapp_options_t opts;
 static struct timespec t_zero;
 static pthread_barrier_t threads_barrier;
+static pthread_mutex_t joining_mutex;
 
 static ftrace_data_t ft_data = {
 	.debugfs = "/sys/kernel/debug",
@@ -448,16 +450,9 @@ int run(int ind,
 	return perf;
 }
 
-static void
-shutdown(int sig)
+static void wakeup_all_threads(void)
 {
 	int i;
-
-	if(!continue_running)
-		return;
-
-	/* notify threads, join them, then exit */
-	continue_running = 0;
 
 	/* Force wake up of all waiting threads */
 	for (i = 0; i <  opts.nresources; i++) {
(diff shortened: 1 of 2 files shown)


src/rt-app.c
```diff
@@ -469,10 +470,24 @@ shutdown(int sig)
...
 
+	/*
+	 * If we try to interrupt pthread_join() at the end of main then we
+	 * might end up joining the same thread twice which is undefined
+	 * behaviour and can lead to segfaults.
+	 *
+	 * Only join if nowhere else started the join process already.
+	 */
+	if (joining)
+		return;
```
jlelli: Mmm, maybe I'm just being overly cautious, but can't this be racy w.r.t. a sigint happening while main is just after this point? shutdown from signal could see joining still 0, pass this check and join threads again?

qais-yousef: Yes but FWIW that's adequate enough IMHO. But I just realized though that I set continue_running *after* wakeup_all_threads() which is wrong. So I'll change this to a mutex with trylock since I have to touch the code again anyway :)

qais-yousef:
Is there anything I need to do for this to be merged?

src/rt-app.c
```diff
@@ -469,19 +463,30 @@ shutdown(int sig)
...
-	 * behaviour and can lead to segfaults.
+	 * Make sure pthread_join() is done once or we hit undefined behaviour
+	 * of multiple simulataneous calls to pthread_join().
 	 *
-	 * Only join if nowhere else started the join process already.
+	 * This is a problem because __shutdown() could be called
+	 * asynchronously from some signals and from main().
 	 */
-	if (joining)
+	if (pthread_mutex_trylock(&joining_mutex))
```
jlelli: Wonder if it's OK that owner doesn't release the mutex before terminating.
I feel uncomfortable about it. :-)

qais-yousef: I chose not to release it because it should be held until we exit, and IMHO an implicit release by exit is the more accurate one. But FWIW, it shouldn't really matter.

I'll create a new PR with this requested change applied. Is this the only remaining issue for this to get merged? If not, can you please aggregate them all in one go so that I'll minimize my work :-)

jlelli: > an implicit release by exit

Is this documented somewhere?
Couldn't find that this is indeed the expected behaviour anywhere. :-/

qais-yousef: On call to exit() we terminate the application along with all of its resources. That's what I meant by implicitly releasing the mutex. The call to unlock() before exit() doesn't add anything to correctness, in contrary one could argue a potential a race between releasing the lock and calling exit() - a late ctrl+c before we just call exit() for instance?

vingu-linaro:
Hi Juri and Qais,

I have just catch up with all your reviews and comments and the changes look fine to me.
I agree with Juri that the 2 last commit should be squash in one.

Thanks for fixing this

src/rt-app.c
jlelli: Guess this is true.
I don't think I have any other comment to this PR (apart from if you could please squash).
Thanks!

qais-yousef: I'll add a comment here anyway and I'll do the squash. Thanks for the review :)

qais-yousef:
Thanks for all the reviews. I have updated the PR and hopefully it's up to the required standard now. Please let me know if the new comment is unclear or the updated squashed commit message needs revising.

jlelli:
Awesome!
Thanks a lot for bearing with the review process. ;-)

PR #68, opened 2018-10-05
permit strings as log level parameters

v0lker:


src/rt-app_utils.c
```diff
@@ -321,3 +332,22 @@ void ftrace_write(int mark_fd, const char *fmt, ...)
...
+int
+str2loglevel(char *s)
+{
+	for (int n=0; lls[n].val != LOG_LEVEL_END; n++) {
+		int equals = 1;
+		int done = 0;
+		for (char *hay = lls[n].s, *ndl = s; !done && equals; hay++, ndl++) {
+			equals = (toupper(*ndl) == toupper(*hay));
+			done = (! *ndl) && (! *hay);
+		}
```
vingu-linaro: Why do you create your own string comparator instead of using strncmp ?

v0lker: because i need to compare the uppercase variant of the string(s). would you prefer it to copy the string while converting to uppercase first, then strcmp()ing them?

vingu-linaro: 1st of all do we really want to support both upper and lower case ? TBH i'm not sure
Then, if we really want to support both lower and upper case, why not adding directly both entries in the table ?

v0lker: 1) i thought it should be 'debug', not 'DEBUG' purely because it reads nicer. but if you look at the help string, that already says 'DEBUG', so maybe that's fine?

2) yes.. could duplicate every entry and then convert during init, but that seems a bit much doesn't it? don't know of a way of doing that with the preprocessor, otherwise i'd have done it that way in the beginning. do you?

ALTERNATIVELY, and this is maybe the most straightforward solution: what about (LOG_LEVEL_debug) and no conversion needed? the interface in util.h are the macros (log_debug() etc) anyways...?

up to you...

src/rt-app_utils.h
```diff
@@ -32,11 +32,21 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
...
-#define LOG_LEVEL_CRITICAL 10
+#define LLS \
+	C(DEBUG, 100) \
+	C(INFO, 75) \
+	C(NOTICE, 50) \
+	C(ERROR, 10) \
+	C(CRITICAL, 10) \
+	C(END, 0)
+
+#define C(x, v) LOG_LEVEL_ ## x = v,
```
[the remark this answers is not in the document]
v0lker: it has to have the same name - it doesn't work otherwise.

src/rt-app_utils.c
v0lker: ok.

src/rt-app_utils.c
```diff
@@ -321,3 +332,15 @@ void ftrace_write(int mark_fd, const char *fmt, ...)
 	}
 
 }
+
+int
+str2loglevel(char *s)
+{
+	for (int n=0; lls[n].val != LOG_LEVEL_END; n++) {
+		if (!strcmp(lls[n].s, s)) {
```
vingu-linaro: always prefer strncmp to strcmp. strcmp relies on null terminating string which is often seen as a weakness
strncmp(lls[n].s, s, strlenght(lls[n].s))

src/rt-app_utils.c
```diff
@@ -31,6 +32,16 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
...
 int log_level = 50;   // default
 
+#define C(x, v) { #x, LOG_LEVEL_ ## x},
+struct loglevel {
+	char *s;
+	int val;
+};
+
+struct loglevel lls[] = { LLS };
+#undef C
```
vingu-linaro: I still don't like this undef then re-def. This is quite confusing and over complexify the code for no good reason. Furthermore, the "levels" are defined at 2 different places so you can't easily know what is it.
Gather everything related to level in one place and don't use undef

src/rt-app_utils.c
v0lker: yes, for comparisons with user input, but here it seems a bit pointless - how would you end up with a string literal that isn't null-terminated? am i missing something?

src/rt-app_utils.c
v0lker: i don't understand what you mean by that the levels are defined in 2 different places? clearly this is exactly the advantage of the above approach that you don't have to maintain 2 separate definitions, no? how would you do that without re-using one macro, i wonder...

src/rt-app_utils.c
vingu-linaro: fair enough

src/rt-app_args.c
```diff
@@ -43,8 +43,8 @@ char help_full[] = \
 "through the standard input.\n\n"
 "Miscellaneous:\n"
 "  -v, --version      display version information and exit\n"
-"  -l, --log          set verbosity level (10: ERROR/CRITICAL, 50: NOTICE (default)\n"
-"                                    75: INFO, 100: DEBUG)\n"
+"  -l, --log          set verbosity level (10: error/critical, 50: notice (default)\n"
```
vingu-linaro: use either error or critical but error/critical is pointless and can be confusing: we can wonder if we have to write "error", "critical" or "error/critical"

v0lker: yes, i was wondering about this as well but decided to keep the original values. there is some confusion with your proposed solution, it wouldn't be clear which level is what, though.

how about `(10: critical, 20: error, 50: notice (default) ...)` ? then it would be clear what the string should be and what the values are?

v0lker: ping...

vingu-linaro: my goal is to remove the level associated to the string from the help. if you use the string, you don't care about the value but the functional level

v0lker: as in `(critical, error, notice (default), ....)`, just without the numbers? yes, could remove them, if they bother you. could you please confirm?

i assume 'error' should still get a new value, as proposed above - ?

vingu-linaro: yes for both questions.

v0lker: ok

PR #73, opened 2018-11-28
rt-app: Add support for SCHED_IDLE policy

vireshk:
This adds support for SCHED_IDLE scheduling policy.

This locally defines SCHED_IDLE macro, which is incorrect but build was
failing without it and I am not sure how to fix that. The arch specific
sched.h headers (like /usr/include/aarch64-linux-gnu/bits/sched.h)
define them only if __USE_GNU is defined.

Signed-off-by: Viresh Kumar <viresh.kumar@linaro.org>

### DIFF
--- a/src/rt-app_types.h
+++ b/src/rt-app_types.h
@@ -42,11 +42,17 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #define EXIT_INV_CONFIG 2
 #define EXIT_INV_COMMANDLINE 3
 
+/* SCHED_IDLE is not available if __USE_GNU is not defined */
+#ifndef __USE_GNU
+#define SCHED_IDLE 5
+#endif
+
 struct _thread_data_t;
 
 typedef enum policy_t
 {
 	other = SCHED_OTHER,
+	idle = SCHED_IDLE,
 	rr = SCHED_RR,
 	fifo = SCHED_FIFO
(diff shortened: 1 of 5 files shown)


src/rt-app_types.h
```diff
@@ -42,9 +42,13 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #define EXIT_INV_CONFIG 2
 #define EXIT_INV_COMMANDLINE 3
 
+/* It is not available in headers without defining __USE_GNU macro */
+#define SCHED_IDLE 5
```
vingu-linaro: can you add a #ifndef __USE_GNU to prevent multiple definition ?

vireshk: Okay,.

vireshk: Vincent, I have made the change you were looking for. Thanks.

vingu-linaro:
looks good to me

src/rt-app_types.h
jlelli: Could you rephrase to something like "SCHED_IDLE is not available if __USE_GNU is not defined" ?
Also, I'd update changelog, since we now have a reasonable solution to this problem. :-)

vireshk:
Any updates on this @vingu-linaro  @jlelli  ?

jlelli:
Thanks Viresh!

PR #76, opened 2019-02-22
Fix unique timer multi instance

vingu-linaro:
Fixes a regression introduced with fork event.
The unique timer were no more per instance but shared between all instance

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -109,6 +109,25 @@ static void thread_data_set_unique_name(thread_data_t *tdata, int nforks)
 }
 
 /*
+ * Give each created thread a list of unique resources.
+ */
+static int thread_data_create_unique_resources(thread_data_t *tdata, const thread_data_t *td)
+{
+	rtapp_resources_t *table = td->local_resources;
+	int local_size = sizeof(rtapp_resources_t) + sizeof(rtapp_resource_t) * table->nresources;
+
+	tdata->local_resources = malloc(local_size);
+	if (!tdata->local_resources) {
+		log_error("Failed to duplicate thread local resources: %s", td->name);
+		return -1;
+	}
+
+	memcpy(tdata->local_resources, td->local_resources, local_size);
+
+	return 0;
+}
+
+/*
  * Create a new pthread using the info provided in the passed @td thread_data_t.
  *
  * @td:		The thread data of the task to create - we copy it and leave
@@ -155,6 +174,10 @@ static int create_thread(const thread_data_t *td, int index, int forked, int nfo
 
 	/* Make sure each (forked) thread has a unique name */
 	thread_data_set_unique_name(tdata, nforks);
+
+	/* Make sure each (forked) thread has its own unique resources */
+	if(thread_data_create_unique_resources(tdata, td))
+		return -1;
 
 	setup_thread_logging(tdata);
 
@@ -365,11 +388,13 @@ static void memload(unsigned long count, struct _rtapp_iomem_buf *iomem)
 }
 
 static int run_event(event_data_t *event, int dry_run,
-		unsigned long *perf, rtapp_resource_t *resources,
+		unsigned long *perf, thread_data_t *tdata,
 		struct timespec *t_first, log_data_t *ldata)
 {
-	rtapp_resource_t *rdata = &(resources[event->res]);
-	rtapp_resource_t *ddata = &(resources[event->dep]);
+	/* By default we use global resources */
+	rtapp_resources_t *table_resources = *(tdata->global_resources);
+	rtapp_resource_t *rdata = &(table_resources->resources[event->res]);
+	rtapp_resource_t *ddata = &(table_resources->resources[event->dep]);
 	unsigned long lock = 0;
 
 	switch(event->type) {
@@ -461,6 +486,11 @@ static int run_event(event_data_t *event, int dry_run,
 			ldata->duration += timespec_to_usec(&t_end);
 		}
 		break;
+	case rtapp_timer_unique:
+		{
+			/* Unique timer uses local resources */
+			rdata = &(tdata->local_resources->resources[event->res]);
+		}
 	case rtapp_timer:
 		{
 			struct timespec t_period, t_now, t_wu, t_slack;
@@ -588,13 +618,13 @@ static int run_event(event_data_t *event, int dry_run,
 	return lock;
 }
 
-int run(int ind,
+int run(thread_data_t *tdata,
 	phase_data_t *pdata,
-	rtapp_resource_t *resources,
 	struct timespec *t_first,
 	log_data_t *ldata)
 {
 	event_data_t *events = pdata->events;
+	int ind = tdata->ind;
 	int nbevents = pdata->nbevents;
 	int i, lock = 0;
 	unsigned long perf = 0;
@@ -610,7 +640,7 @@ int run(int ind,
 						"[%d] executing %d",
 						ind, i);
 		lock += run_event(&events[i], !continue_running, &perf,
-				  resources, t_first, ldata);
+				  tdata, t_first, ldata);
 	}
 
 	return perf;
@@ -619,14 +649,20 @@ int run(int ind,
 static void wakeup_all_threads(void)
 {
 	int i;
+	int nresources = opts.resources->nresources;
+	rtapp_resource_t *resources = opts.resources->resources;
 
-	/* Force wake up of all waiting threads */
-	for (i = 0; i <  opts.nresources; i++) {
-		if (opts.resources[i].type == rtapp_wait) {
-			pthread_cond_broadcast(&opts.resources[i].res.cond.obj);
+	/*
+	 * Force wake up of all waiting threads.
+	 * At now we don't need to look into local resources because rtapp_wait
+	 * and rtapp_barrier are always global
+	 */
+	for (i = 0; i <  nresources; i++) {
+		if (resources[i].type == rtapp_wait) {
+			pthread_cond_broadcast(&resources[i].res.cond.obj);
 		}
-		if (opts.resources[i].type == rtapp_barrier) {
-			pthread_cond_broadcast(&opts.resources[i].res.barrier.c_obj);
+		if (resources[i].type == rtapp_barrier) {
+			pthread_cond_broadcast(&resources[i].res.barrier.c_obj);
 		}
 	}
 }
@@ -1056,8 +1092,7 @@ void *thread_body(void *arg)
 
 		memset(&ldata, 0, sizeof(ldata));
 		clock_gettime(CLOCK_MONOTONIC, &t_start);
-		ldata.perf = run(data->ind, pdata, *(data->resources),
-				&t_first, &ldata);
+		ldata.perf = run(data, pdata, &t_first, &ldata);
 		clock_gettime(CLOCK_MONOTONIC, &t_end);
 
 		if (timings)
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -495,18 +495,18 @@ static int run_event(event_data_t *event, int dry_run,
 		break;
 	case rtapp_suspend:
 		{
-		log_debug("suspend %s ", rdata->name);
-		pthread_mutex_lock(&(ddata->res.mtx.obj));
-		pthread_cond_wait(&(rdata->res.cond.obj), &(ddata->res.mtx.obj));
-		pthread_mutex_unlock(&(ddata->res.mtx.obj));
+			log_debug("suspend %s ", rdata->name);
+			pthread_mutex_lock(&(ddata->res.mtx.obj));
+			pthread_cond_wait(&(rdata->res.cond.obj), &(ddata->res.mtx.obj));
+			pthread_mutex_unlock(&(ddata->res.mtx.obj));
 		break;
 		}
 	case rtapp_resume:
 		{
-		log_debug("resume %s ", rdata->name);
-		pthread_mutex_lock(&(ddata->res.mtx.obj));
-		pthread_cond_broadcast(&(rdata->res.cond.obj));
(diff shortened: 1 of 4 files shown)


src/rt-app_parse_config.c
```diff
@@ -325,40 +335,50 @@ static void
...
 	if (json_object_is_type(resources, json_type_object)) {
-		opts->nresources = 0;
+		opts->resources->nresources = 0;
 		foreach(resources, entry, key, val, idx) {
-			opts->nresources++;
+			opts->resources->nresources++;
 		}
 
-		log_info(PFX "Found %d Resources", opts->nresources);
-		opts->resources = malloc(sizeof(rtapp_resource_t) * opts->nresources);
```
qais-yousef: Sigh. I see the code already doesn't do that. Looking at the code we have a few of these actually. Oh well.

src/rt-app_parse_config.c
```diff
@@ -325,40 +335,50 @@ static void
...
+		log_info(PFX "Found %d Resources", opts->resources->nresources);
+		size = sizeof(rtapp_resources_t) + sizeof(rtapp_resource_t) * opts->resources->nresources;
 
+		opts->resources = realloc(opts->resources, size);
```
qais-yousef: Validate the return value?

src/rt-app_parse_config.c
```diff
@@ -830,14 +853,28 @@ parse_task_phase_data(struct json_object *obj,
...
 	struct json_object *phases_obj, *resources;
 
 	log_info(PFX "Parsing task %s [%d]", name, index);
 
 	/* common and defaults */
-	data->resources = &opts->resources;
+	data->opts = opts;
+
+	/*
+	 * Set a pointer to opt entry as we miht end up to modify the localtion of
```
qais-yousef: s/miht/might/

src/rt-app_parse_config.c
```diff
@@ -830,14 +853,28 @@ parse_task_phase_data(struct json_object *obj,
...
+	 * global resources during realloc
+	 */
+	data->global_resources = &(opts->resources);
+
+	/*
+	 * Create at least a "empty" struct that will then be filled either will
+	 * parsing resources table or when parsing task's event
+	 */
+	data->local_resources = malloc(sizeof(rtapp_resources_t));
```
qais-yousef: Check returned value is valid

src/rt-app_types.h
```diff
@@ -206,6 +212,10 @@ typedef struct _thread_data_t {
 
 	int forked;
 	int num_instances;
+
+	struct _rtapp_options_t *opts;
+	rtapp_resources_t *local_resources;
+	rtapp_resources_t **global_resources;
```
qais-yousef: Does gloabal_resources need to be a double pointer? It seems to me an artefact left over from previous code?

src/rt-app_parse_config.c
```diff
@@ -830,14 +853,28 @@ parse_task_phase_data(struct json_object *obj,
 
 static void
 parse_task_data(char *name, struct json_object *obj, int index,
-		  thread_data_t *data, rtapp_options_t *opts)
+			thread_data_t *data, rtapp_options_t *opts)
```
qais-yousef: Indentation change? It looks broken in this view at least - but if it looks OK in a proper text editor then ignore me.

src/rt-app.c
```diff
@@ -108,6 +108,26 @@ static void thread_data_set_unique_name(thread_data_t *tdata, int nforks)
...
+/*
+ * Give each created thread a list of unique resources.
+ */
+static int thread_data_create_unique_resources(thread_data_t *tdata, const thread_data_t *td)
+{
+	rtapp_resources_t *table = td->local_resources;
+	int local_size = sizeof(rtapp_resources_t) + sizeof(rtapp_resource_t) * table->nresources;
+
+	tdata->local_resources = malloc(local_size);
+	//log_info(PIN2 "thread_data_create_unique_resources %x size %d", tdata->local_resources, size);
```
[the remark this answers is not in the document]
vingu-linaro: yes. will remove

src/rt-app.c
```diff
@@ -619,14 +648,20 @@ int run(int ind,
...
-	for (i = 0; i <  opts.nresources; i++) {
-		if (opts.resources[i].type == rtapp_wait) {
-			pthread_cond_broadcast(&opts.resources[i].res.cond.obj);
+	/*
+	 * Force wake up of all waiting threads.
+	 * At now we don't need to look into local resources because rtapp_wait
+	 * and rtapp_barrier are always global
+	 */
+	for (i = 0; i <  nresources; i++) {
+		if (resources[i].type == rtapp_wait) {
```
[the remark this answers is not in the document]
vingu-linaro: maybe but let consider that in a dedicated clean up pul request

src/rt-app_parse_config.c
```diff
@@ -301,22 +303,30 @@ parse_resource_data(const char *name, struct json_object *obj, int idx,
...
 	log_info(PIN2 "Add new resource %s [%d] type %d", name, idx, type);
 
-	opts->nresources++;
-	opts->resources = realloc(opts->resources, sizeof(rtapp_resource_t) * opts->nresources);
+	table->nresources++;
 
-	init_resource_data(name, type, idx, opts);
+	size = sizeof(rtapp_resources_t) + sizeof(rtapp_resource_t) * table->nresources;
+
+	*resources_table = realloc(*resources_table, size);
```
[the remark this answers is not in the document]
vingu-linaro: yes. I have to add the error path
more generally speaking, we need to check all error path

src/rt-app_parse_config.c
```diff
@@ -325,40 +335,50 @@ static void
...
 	struct lh_entry *entry; char *key; struct json_object *val; int idx;
+	int size;
 
 	log_info(PFX "Parsing resource section");
 
+	/*
+	 * Create at least a "empty" struct that will then be filled either will
+	 * parsing resources table or when parsing task's event
+	 */
+	opts->resources = malloc(sizeof(rtapp_resources_t));
```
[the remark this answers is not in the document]
vingu-linaro: yes

src/rt-app_parse_config.c
vingu-linaro: I will check

src/rt-app_parse_config.c
vingu-linaro: yes

src/rt-app_types.h
vingu-linaro: yes it has to be double pointer
because realloc can modify the pointer, we must be able to save the new pointer.

qais-yousef: Maybe it's friday afternoon that's clouding my thinking. But we can save the returned pointer with a single pointer too. What am I missing? When forking we use realloc to expand `threads` structure - which is a single pointer. If we fail to expand then we basically fail to run the whole test and bail out.

vingu-linaro: realloc can move the  struct somewhere else so we must update the place where the pointer is saved. In fact, this happened during some tests.
If you look at add_resource_data(), we don't use the local pointer table but *resources_table after the realloc

qais-yousef: >If you look at add_resource_data(), we don't use the local pointer table but *resources_table after the realloc

Yes because it's a double pointer. If you make it a single pointer you can just access it directly, no?

Because you added rtapp_resources_t there's no need for the double pointer AFAICS and I don't see why local_resources can be a single pointer but not global_resources since the table is now embedded in the new struct.

We expand threads in fork event in rt-app.c using realloc, why the usage of realloc in the 2 cases is not equivalent?

vingu-linaro: no, otherwise we would only update the "local" pointer of gloabl_resources of the task that has added the resources but other tasks will continue to point to previous one.
single pointer is enough for loacl_resources because it's not point by others.

we have opts.resources->the_global_resouces and threadA_data.global_resources-> opts.resources->the_global_resouces and threadB_data.global_resources-> opts.resources->the_global_resouces. If threadA_data and threadB_data were directly pointing to the_global_resouces instead, if @ of the_global_resouces changes, only the thread that did the changes would have the new @ others would still point to previous @

qais-yousef: Oh I see what I was missing now. This is a per thread pointer to a global resource. So we are saving a pointer to a pointer actually - I was thinking the table stored directly in this struct and missed it's part of the tread data, my bad.

If we already have a pointer to the global_resources in opts then we should just use that IMHO. It's very easy to create a micro to improve readability. But I don't feel strongly about it.

vingu-linaro: This pointer is used quite often when running events and make the code more readable instead of finding our way to the pointer every time.

vingu-linaro: opts pointer is now removed from thread structure because it was only used for resources init like pi_enable

vingu-linaro:
I have updated the fix and remove the finally useless opts pointer.
I have also checked the returned value of meme allocation

vingu-linaro:
Have you got comments ? otherwise i'm going to merge it

qais-yousef:
The change looks fine to me. I don't know if @jlelli wants to add anything.

It would have been nice to split the error checking cleanups into a separate patch or update the commit message to mention this cleanup at least.

src/rt-app.c
```diff
@@ -495,18 +525,18 @@ static int run_event(event_data_t *event, int dry_run,
 		break;
 	case rtapp_suspend:
 		{
-		log_debug("suspend %s ", rdata->name);
-		pthread_mutex_lock(&(ddata->res.mtx.obj));
-		pthread_cond_wait(&(rdata->res.cond.obj), &(ddata->res.mtx.obj));
-		pthread_mutex_unlock(&(ddata->res.mtx.obj));
+			log_debug("suspend %s ", rdata->name);
```
jlelli: This and right below seems unrelated to this change?

src/rt-app.c
```diff
@@ -712,6 +748,11 @@ static void create_cpuset_str(cpuset_data_t *cpu_data)
 	}
 
 	cpu_data->cpuset_str = malloc(size_needed);
+	if (!cpu_data->cpuset_str) {
```
jlelli: This unrelated as well?

src/rt-app.c
```diff
@@ -952,6 +995,10 @@ void *thread_body(void *arg)
 	/* Init timing buffer */
 	if (opts.logsize > 0) {
 		timings = malloc(opts.logsize);
+		/*
```
jlelli: Unrelated?

src/rt-app.c
```diff
@@ -1177,6 +1223,10 @@ int main(int argc, char* argv[])
 	/* allocated threads */
 	nthreads = opts.nthreads;
 	threads = malloc(nthreads * sizeof(pthread_t));
+	if (!threads) {
```
jlelli: Unrelated?

src/rt-app_parse_config.c
```diff
@@ -526,27 +561,27 @@ parse_task_event_data(char *name, struct json_object *obj,
 	if (!strncmp(name, "timer", strlen("timer"))) {
 
 		tmp = get_string_value_from(obj, "ref", TRUE, "unknown");
-		if (!strncmp(tmp, "unique", strlen("unique")))
-				ref = create_unique_name(unique_name, sizeof(unique_name), tmp, tag);
-		else
-				ref = tmp;
 
-		i = get_resource_index(ref, rtapp_timer, opts);
+		if (!strncmp(tmp, "unique", strlen("unique"))) {
```
jlelli: Mmm. I'm wondering if for forking tasks timers should be unique by default?

src/rt-app_parse_config.c
```diff
@@ -526,27 +561,27 @@ parse_task_event_data(char *name, struct json_object *obj,
...
+			ref = create_unique_name(unique_name, sizeof(unique_name), tmp, tag);
+			resources_table = &tdata->local_resources;
+			data->type = rtapp_timer_unique;
```
jlelli: Do we need a different type? Can't we make timers local even for tasks that don't fork?

src/rt-app.c
vingu-linaro: the 1st commit is doing a cleanup 
the 2nd fix the unique timer problem

src/rt-app.c
vingu-linaro: While fixing unique timer, missing error path has been highlighted. All this are in the 1st patch

src/rt-app_parse_config.c
vingu-linaro: I don't think. I have in mind some UC where threads fork but still use the same timer

src/rt-app_parse_config.c
vingu-linaro: not sure to get your point. "unique" is already a key of rt-app timer description
Any timer with a "unique" name will be unique/local to each thread whatever the thread is created at the beg or forked later.

src/rt-app_parse_config.c
jlelli: ok

src/rt-app_parse_config.c
jlelli: You actually answered my doubt in question above.. there are cases where tasks need to share timers.. so we need that distinction.

vingu-linaro:
I have updated patchset:
-  No code change
- move some cleanup changes that were in patch2 into patch1.

jlelli:
LGTM, tested as well.
I'm going to merge this, thanks a lot @vingu-linaro .
One last thing that I was thinking however is that it would be nice to document how we handle resources (data structures, etc.). Do you think you could add some of that somewhere as a comment?

vingu-linaro:
I'm going to prepare a patch to add such details

jlelli:
Awesome!

PR #70, opened 2018-10-15
Add taskgroups support for SCHED_OTHER tasks

deggeman:
The taskgroups support works currently for cgroupv1 (legacy)
hierarchies. It can be optionally configured by specifying
--with-taskgroups as a build parameter (see README.in).
A configuration example is given in doc/tutorial.txt.
For now an rt-app binary w/ taskgroup support only allows SCHED_NORMAL
policy. An rt-app binary w/o taskgroups support ignores any taskgroup
related name/value pairs in the json file.

Tested on x86_64 and arm64.

Signed-off-by: Dietmar Eggemann <dietmar.eggemann@arm.com>

PR #77, opened 2019-04-10
GitHub deg/taskgroups sched other v4

deggeman:


deggeman:
I close this one #77 since I sent the update against #70 instead.

PR #79, opened 2019-05-23
Couple of trivial patches

claudioscordino:
- Fix an annoying segfault when logdir is provided but not existing.
- Small changes to the existing documentation.


vingu-linaro:
looks good to me

PR #80, opened 2019-06-13
Improve FTrace support and make it LISA compliant

derkling:
This will give LISA a chance to use ftrace events to know about rt-app tasks and write tests without the need to parse logfiles and try hard (not always successfully) to align logfile timestamps with ftrace timestamps.


src/rt-app.c
```diff
@@ -639,10 +639,9 @@ int run(thread_data_t *tdata,
 			return perf;
 
 		log_debug("[%d] runs events %d type %d ", ind, i, events[i].type);
-		if (opts.ftrace)
-				log_ftrace(ft_data.marker_fd,
-						"[%d] executing %d",
-						ind, i);
+		log_ftrace(ft_data.marker_fd, FTRACE_EVENT,
+			   "rtapp_event: id=%d type=%d desc=%s",
```
jlelli: It would be nice to also print duration for some of the event types (e.g., run, sleep, timer).
Wonder if that clutters code too much, though?

doc/tutorial.txt
```diff
@@ -75,8 +75,13 @@ The use of a temporary buffer prevents the threads of unexpected wait during
 io access. The "Auto" mode is not implemented yet and fallback to "file" mode
 for the moment.
 
-* ftrace: Boolean. If enabled, rt-app logs in ftrace the main events of the use
-case. Default value is False.
+* ftrace: String. If not "non", rt-app logs in ftrace the events corresponding
```
jlelli: s/non/none/

doc/tutorial.txt
```diff
@@ -598,6 +598,13 @@ The generated events are of these main categories:
 
   Reporting the execution of the specified event type in the current phase/loop.
 
+* Workload's stats, for example:
```
jlelli: Please add "- stats" in the corresponding "ftrace" field section (above?).

src/rt-app.c
```diff
@@ -1128,6 +1128,14 @@ void *thread_body(void *arg)
 		log_ftrace(ft_data.marker_fd, FTRACE_LOOP,
 			   "rtapp_loop: event=end thread_loop=%d phase=%d phase_loop=%d",
 			   thread_loop, phase, phase_loop);
+		log_ftrace(ft_data.marker_fd, FTRACE_STATS,
+			   "rtapp_stats: period=%d run=%d wu_lat=%d slack=%d c_run=%d c_period=%d",
+			   curr_timing->period,
+			   curr_timing->duration,
+			   curr_timing->wu_latency,
+			   curr_timing->slack,
+			   curr_timing->c_period,
```
jlelli: c_period and c_duration are inverted (w.r.t. print format).

jlelli:
Nice work! I like it.
Apart from the few things I pointed out, I was thinking that we might want to remove explicit mentioning of LISA in changelogs, subjects, comments, etc.
Not that I don't like LISA (:-)), but the changes look more general to me, and good even if LISA is not used.
Thanks!

src/rt-app.c
derkling: I guess we can add one category to track this more detailed info.
However, what we have so far with the current set of events, it's good enough to compute the duration of each event in post-processing. We have indeed the trace timestamp and a start/stop marker for each command.

src/rt-app.c
derkling: Good catch!

derkling:
@vlegout any major objection for this series?

doc/tutorial.txt
```diff
@@ -75,8 +75,18 @@ The use of a temporary buffer prevents the threads of unexpected wait during
...
+  measured after the completion of each task activation.
+
+* ftrace: String. If not "none", rt-app logs in ftrace the events corresponding
+to the requested categories. Supported categories are:
+ - main  : events generated by the main rt-app task
+ - task  : events generated by a task
+ - run   : events generated by the run of a task
+ - loop  : events generated by each loop of a task
+ - stats : events reporting statistics on task's activation
+Default value is "none".
```
vingu-linaro: you should keep some backward compatibility
please still mention true/false as deprecated

doc/tutorial.txt
```diff
@@ -75,6 +75,10 @@ The use of a temporary buffer prevents the threads of unexpected wait during
 io access. The "Auto" mode is not implemented yet and fallback to "file" mode
 for the moment.
 
+* logstats: Boolean. Report tasks generated statistics into a logfile. A logfile
+  will be generated for each task, within the logdir, reporting the main metrics
```
vingu-linaro: we already have "logsize": "disable" for that prupose

src/rt-app.c
```diff
@@ -1212,6 +1215,9 @@ void setup_thread_logging(thread_data_t *tdata)
 	tdata->main_app_start = t_start;
 	tdata->lock_pages = opts.lock_pages;
 
+	if (!opts.logstats)
+		return;
```
vingu-linaro: this one is a fix for not creating empty logfile but use  opts.logsize instead

doc/tutorial.txt
vingu-linaro: and True should enable all traces

doc/tutorial.txt
derkling: What a horrible naming choice for that parameter! :)

Anyway, it's kind of broken since also with the `disable` option we still generate a logfile, it's just empty :)

Anyway, I guess we still want to go for backward compatibility, thus I will keep this changeset minimal by just fix the empty file generation.

Let's keep instead of the (backward compatible) refactoring of the parameter sematic for the future. ;)

derkling:
@vingu-linaro I think I've addressed all your comments.
Any chance to follow up on this soonish?

src/rt-app.c
```diff
@@ -1174,7 +1180,7 @@ void *thread_body(void *arg)
 	/* Force thread into root taskgroup. */
 	reset_thread_taskgroup();
 
-	if (timings) {
+	if (opts.logsize && timings) {
```
vingu-linaro: Why have you added the test on opts.logsize ?
timings is null if logsize is <= 0

vingu-linaro:
Apart from the comment on the reason for adding a test on logsize when flushing the buffer in the file, looks good to me

derkling:
Ok, fixed the last comment from @vingu-linaro, thanks.

This should unlock @jlelli for merge, isn't it? ;)

PR #81, opened 2019-07-08
rt-app: add uclamp support

derkling:
Utilization clamping allows to set task-specific utilization clamp
values, for SCHED_OTHER and SCHED_{FIFO,RR}, via the sched_setattr(2)
syscall.

Core bits and per-task API has been added to the Linux kernel in version v5.3:
   https://lore.kernel.org/lkml/20190621084217.8167-1-patrick.bellasi@arm.com/

Signed-off-by: Patrick Bellasi <patrick.bellasi@arm.com>

### DIFF
--- a/libdl/dl_syscalls.h
+++ b/libdl/dl_syscalls.h
@@ -74,22 +74,49 @@
 #define RLIMIT_DLDLINE		16
 #define RLIMIT_DLRTIME		17
 
+/*
+ * For the sched_{set,get}attr() calls
+ */
+#define SCHED_FLAG_RESET_ON_FORK        0x01
+#define SCHED_FLAG_RECLAIM              0x02
+#define SCHED_FLAG_DL_OVERRUN           0x04
+#define SCHED_FLAG_KEEP_POLICY          0x08
+#define SCHED_FLAG_KEEP_PARAMS          0x10
+#define SCHED_FLAG_UTIL_CLAMP_MIN       0x20
+#define SCHED_FLAG_UTIL_CLAMP_MAX       0x40
+
+#define SCHED_FLAG_KEEP_ALL 	(SCHED_FLAG_KEEP_POLICY | \
+				 SCHED_FLAG_KEEP_PARAMS)
+
+#define SCHED_FLAG_UTIL_CLAMP	(SCHED_FLAG_UTIL_CLAMP_MIN | \
+				 SCHED_FLAG_UTIL_CLAMP_MAX)
+
+#define SCHED_FLAG_ALL	(SCHED_FLAG_RESET_ON_FORK	| \
+			 SCHED_FLAG_RECLAIM		| \
+			 SCHED_FLAG_DL_OVERRUN		| \
+			 SCHED_FLAG_KEEP_ALL		| \
+			 SCHED_FLAG_UTIL_CLAMP)
+
 struct sched_attr {
 	__u32 size;
-	
+
 	__u32 sched_policy;
 	__u64 sched_flags;
-	
+
 	/* SCHED_NORMAL, SCHED_BATCH */
 	__s32 sched_nice;
-	
+
 	/* SCHED_FIFO, SCHED_RR */
 	__u32 sched_priority;
-	
+
 	/* SCHED_DEADLINE */
 	__u64 sched_runtime;
 	__u64 sched_deadline;
 	__u64 sched_period;
+
+	/* Utilization hints */
+	__u32 sched_util_min;
+	__u32 sched_util_max;
 };
 
 int sched_setattr(pid_t pid,
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -1010,12 +1010,18 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 			sa_params.sched_flags |= SCHED_FLAG_UTIL_CLAMP_MIN;
 			log_notice("[%d] setting util_min=%d",
 				   data->ind, sched_data->util_min);
+			log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+				   "rtapp_attrs: event=uclamp util_min=%d",
+				   sched_data->util_min);
 		}
 		if (sched_data->util_max != -1) {
 			sa_params.sched_util_max = sched_data->util_max;
 			sa_params.sched_flags |= SCHED_FLAG_UTIL_CLAMP_MAX;
 			log_notice("[%d] setting util_max=%d",
 				   data->ind, sched_data->util_max);
+			log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+				   "rtapp_attrs: event=uclamp util_max=%d",
+				   sched_data->util_max);
 		}
 
 		ret = sched_setattr(tid, &sa_params, flags);
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -854,7 +854,7 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 	struct sched_param param;
 	policy_t policy;
 #ifdef DLSCHED
-	struct sched_attr dl_params;
+	struct sched_attr sa_params;
 	pid_t tid;
 	unsigned int flags = 0;
 #endif
@@ -944,15 +944,15 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 					sched_data->runtime, sched_data->deadline);
 
 			tid = gettid();
-			dl_params.size = sizeof(struct sched_attr);
-			dl_params.sched_flags = 0;
-			dl_params.sched_policy = SCHED_DEADLINE;
-			dl_params.sched_priority = 0;
-			dl_params.sched_runtime = sched_data->runtime;
-			dl_params.sched_deadline = sched_data->deadline;
-			dl_params.sched_period = sched_data->period;
+			sa_params.size = sizeof(struct sched_attr);
+			sa_params.sched_flags = 0;
+			sa_params.sched_policy = SCHED_DEADLINE;
+			sa_params.sched_priority = 0;
+			sa_params.sched_runtime = sched_data->runtime;
+			sa_params.sched_deadline = sched_data->deadline;
+			sa_params.sched_period = sched_data->period;
 
-			ret = sched_setattr(tid, &dl_params, flags);
+			ret = sched_setattr(tid, &sa_params, flags);
 			if (ret != 0) {
 				log_critical("[%d] sched_setattr "
 						"returned %d", data->ind, ret);
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -878,6 +878,11 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 		case fifo:
 			log_debug("[%d] setting scheduler %s priority %d", data->ind,
 					policy_to_string(sched_data->policy), sched_data->prio);
+			log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+				   "rtapp_attrs: event=policy policy=%s prio=%d",
+				   sched_data->policy == rr ? "rr" : "fifo",
+				   sched_data->prio);
+
 			param.sched_priority = sched_data->prio;
 			ret = pthread_setschedparam(pthread_self(),
 					sched_data->policy,
@@ -895,7 +900,10 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 		case idle:
 			log_debug("[%d] setting scheduler %s priority %d", data->ind,
 					policy_to_string(sched_data->policy), sched_data->prio);
-
+			log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+				   "rtapp_attrs: event=policy policy=%s prio=%d",
+				   sched_data->policy == other ? "other" : "idle",
+				   sched_data->prio);
 
 			if ((sched_data->policy == other) && (sched_data->prio > 19 || sched_data->prio < -20)) {
 				log_critical("[%d] setpriority "
@@ -942,6 +950,10 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 					" period %lu", data->ind,
 					policy_to_string(sched_data->policy), sched_data->period,
 					sched_data->runtime, sched_data->deadline);
+			log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+				   "rtapp_attrs: event=policy policy=dl runtime=%lu deadline=%lu period=%lu",
+				   sched_data->prio, sched_data->runtime,
+				   sched_data->deadline, sched_data->period);
 
 			tid = gettid();
 			sa_params.size = sizeof(struct sched_attr);
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -970,6 +970,34 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 			exit(EXIT_FAILURE);
 	}
 
+	if (sched_data->policy != deadline &&
+	    (sched_data->util_min != -1 || sched_data->util_max != -1)) {
+		sa_params.size = sizeof(struct sched_attr);
+		sa_params.sched_flags = SCHED_FLAG_KEEP_ALL;
+		tid = gettid();
+
+		if (sched_data->util_min != -1) {
+			sa_params.sched_util_min = sched_data->util_min;
+			sa_params.sched_flags |= SCHED_FLAG_UTIL_CLAMP_MIN;
+			log_notice("[%d] setting util_min=%d",
+				   data->ind, sched_data->util_min);
+		}
+		if (sched_data->util_max != -1) {
+			sa_params.sched_util_max = sched_data->util_max;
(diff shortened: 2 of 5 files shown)


configure.ac
```diff
@@ -26,6 +26,14 @@ AC_ARG_WITH([deadline],
 
 AM_CONDITIONAL([SET_DLSCHED], [test x$with_deadline != xno && test !HAVE_SCHED_SETATTR])
 
+AC_ARG_WITH([uclamp],
+	[AS_HELP_STRING([--with-uclamp],
+		[Add support for SCHEDA_UCLAMP])],
+	[AC_DEFINE([UCLAMP], [1], [Define if you have SCHED_UCLAMP support])],
+	[with_uclamp=no])
+
+AM_CONDITIONAL([SET_UCLAMP], [test x$with_uclamp != xno && test !HAVE_SCHED_SETATTR])
```
qais-yousef: I think this (and the with_deadline) options are not necessary. If the kernel doesn't have the support you should get an error and bailout. Why complicate the building process if we can dynamically manage this?

derkling:
Any volunteer for reviewing this one?

src/rt-app_types.h
```diff
@@ -176,6 +176,8 @@ typedef struct _sched_data_t {
 	unsigned long runtime;
 	unsigned long deadline;
 	unsigned long period;
+	int boost;
+	int cap;
```
qais-yousef: ditto

src/rt-app_parse_config.c
```diff
@@ -791,6 +789,21 @@ static sched_data_t *parse_sched_data(struct json_object *obj, int def_policy)
 	tmp_data.period = get_int_value_from(obj, "dl-period", TRUE, tmp_data.runtime);
 	tmp_data.deadline = get_int_value_from(obj, "dl-deadline", TRUE, tmp_data.period);
 
+	/* clamping params (-1: no changes ) */
+	tmp_data.boost = get_int_value_from(obj, "boost", TRUE, -1);
```
qais-yousef: uclamp_max is a better name; it'd be much clearer to keep the 1:1 naming since the user is expected to know they're using uclamp and renaming it to make it look like something else will add confusion

src/rt-app_parse_config.c
```diff
@@ -791,6 +789,21 @@ static sched_data_t *parse_sched_data(struct json_object *obj, int def_policy)
...
+	if (tmp_data.boost != -1 && tmp_data.boost > 1024) {
+		log_critical(PIN2 "Invalid boost %d (>1024)", tmp_data.boost);
+		exit(EXIT_INV_CONFIG);
+	}
+	log_notice(PIN2 "boost: %d", tmp_data.boost);
+
+	tmp_data.cap = get_int_value_from(obj, "cap", TRUE, -1);
```
qais-yousef: ditto

src/rt-app_types.h
derkling: The idea was to expose a more generic concept which, if we have uclamp, translates in `clamp_{min,max}`.
But I get your point and I agree that at this stage is generates more confusion for nothing.

configure.ac
derkling: Done

src/rt-app_types.h
derkling: Done

src/rt-app_parse_config.c
derkling: Done

src/rt-app_parse_config.c
```diff
@@ -836,7 +836,8 @@ static sched_data_t *parse_sched_data(struct json_object *obj, int def_policy)
 	tmp_data.deadline *= 1000;
 
 	/* Check if we found at least one meaningful scheduler parameter */
-	if (tmp_data.prio != -1 || tmp_data.runtime || tmp_data.period || tmp_data.period) {
```
qais-yousef: Haha. We were checking for period twice here :D

So you're actually fixing a typo here. Worth mentioning in the commit message but that's really a minor nit only worth doing if this series gets updated. But not worth updating it only for that.

libdl/dl_syscalls.h
```diff
@@ -74,22 +74,49 @@
...
-	
+
 	/* SCHED_DEADLINE */
 	__u64 sched_runtime;
 	__u64 sched_deadline;
 	__u64 sched_period;
+
+	/* Utilization hints */
+	__u32 sched_util_min;
+	__u32 sched_util_max;
```
qais-yousef: I'm sorry I might be too picky here. I thought in the user API of uclamp for cgroup we use uclamp_min and uclamp_max? I think this naming is okay but if we use a different name somewhere else I think for consistency it might be worth to consider using one name across. I don't mind if it stays like this if you're happy with it and no one else comments on it.

src/rt-app.c
```diff
@@ -970,6 +970,35 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
...
+			log_notice("[%d] setting util_max=%d",
+				   data->ind, sched_data->util_max);
+		}
+
+		ret = sched_setattr(tid, &sa_params, flags);
+		if (ret != 0) {
+			log_critical("[%d] sched_setattr returned %d",
+				     data->ind, ret);
+			errno = ret;
+			perror("sched_setattr");
```
qais-yousef: sched_setattr::uclamp might give a better indication which sched_setattr() operation failed? When setting sched deadline attribute an identical message will be printed so worth expanding the text to distinguish them.

src/rt-app.c
```diff
@@ -942,6 +950,10 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 					" period %lu", data->ind,
 					policy_to_string(sched_data->policy), sched_data->period,
 					sched_data->runtime, sched_data->deadline);
+			log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+				   "rtapp_attrs: event=policy policy=dl runtime=%lu deadline=%lu period=%lu",
+				   sched_data->prio, sched_data->runtime,
+				   sched_data->deadline, sched_data->period);
```
qais-yousef: This has a different signature than above ie: you print more than just policy and prio. Will it create confusion since IIUC the event name will always be rtapp_attrs when viewed in pandas but we can end up with 2 different set of columns depending on the policy we are setting?

src/rt-app.c
```diff
@@ -1010,12 +1010,18 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 			sa_params.sched_flags |= SCHED_FLAG_UTIL_CLAMP_MIN;
 			log_notice("[%d] setting util_min=%d",
 				   data->ind, sched_data->util_min);
+			log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+				   "rtapp_attrs: event=uclamp util_min=%d",
```
qais-yousef: Same here. Not sure if using the same event name will produce something that will be hard to distinguish in pandas when converted to a data frame by Lisa. I'm probably wrong but worth asking :-)

src/rt-app.c
```diff
@@ -970,6 +970,35 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
...
+			sa_params.sched_flags |= SCHED_FLAG_UTIL_CLAMP_MAX;
+			log_notice("[%d] setting util_max=%d",
+				   data->ind, sched_data->util_max);
+		}
+
+		ret = sched_setattr(tid, &sa_params, flags);
+		if (ret != 0) {
+			log_critical("[%d] sched_setattr returned %d",
+				     data->ind, ret);
+			errno = ret;
```
qais-yousef: Sorry I missed this earlier. From the man pages sched_setattr() will return -1 and set errno. I can see that other places in the code have similar logic. Am I missing something or we're effectively setting errno to -1 here?

src/rt-app_parse_config.c
derkling: This chunk is already in a dedicated patch, do you mean that the changelog of that patch is not clear enough?

libdl/dl_syscalls.h
derkling: What you say makes sense, we should keep consistency.
Unfortunately, the cgroup names (`uclamp_{min,max}`) have been requested by Tejun once the core bits were already mainline, hence we (are going to) have now this asymmetry.

Here I'm using the same names used in Linux mainline's `include/uapi/linux/sched.h`.

I would keep rt-app code aligned to whatever we have in kernel-space... although unfortunate.

src/rt-app.c
derkling: Right, I'll extend to be:
```C
perror("sched_setattr: failed to set uclamp value(s)");
```

src/rt-app.c
derkling: Your right, we will end up with multiple columns but that's intentional.

What I'm trying to do is to cover cases where we have only one clamp updated.
Once we merge the rows in Pandas we will end up with `NaN` values for the clamp not updated. That will allow to easily do a forward propagation of any possible previous update to overwrite the `NaN`s.

Otherwise, to always report both values we will have to either track then in rt-app or always read them before writing. Preferences?

src/rt-app.c
derkling: The same response/question above applies.

src/rt-app.c
derkling: No, that's wrong... lemme get rid of that assignment, as you correctly do in #83.

src/rt-app_parse_config.c
qais-yousef: I guess it is clear. Feel free to ignore this.

src/rt-app.c
qais-yousef: No preference. As long as you've thought about it and have a clear idea on how to deal with it I'm happy with it. We can evolve later if the need ever arises.

src/rt-app_parse_config.c
```diff
@@ -836,7 +836,8 @@ static sched_data_t *parse_sched_data(struct json_object *obj, int def_policy)
...
+	if (tmp_data.prio != -1 ||
+	    tmp_data.runtime || tmp_data.period || tmp_data.deadline) {
```
credp: looks like the right thing to do

src/rt-app.c
```diff
@@ -976,6 +976,23 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
...
 		tid = gettid();
 
+		/*
+		 * Currently we cannot use the syscall to set a clamp
+		 * value without specifing a valid policy and priotiry.
+		 * Read the current values and use them for the following
+		 * syscall.
+		 * TODO: fix __sched_setscheduler() to get rid of this!
+		 */
+		ret = sched_getattr(tid, &sa_params,
```
credp: I think sa_params is only present if DL is enabled - I didn't see anything moving the declaration outside the DLSCHED guards.

qais-yousef:
@vingu-linaro @jlelli : Can you spare sometime to review this please? We need rt-app support to start developing tests for util-clamp in Lisa

vingu-linaro:
We have a regression with this branch on the json file parsing.
with this branch :
$ rt-app doc/examples/template.json 
[rt-app] <crit> [json] Error while parsing config
         [json] 
whereas with master:
$ rt-app doc/examples/template.json 
[rt-app] <notice> Calibrate ns per loop
[rt-app] <notice> pLoad = 19ns : calib_cpu 0
[rt-app] <notice> [0] starting thread ...

[rt-app] <notice> [0] Starting with SCHED_OTHER policy with priority 0
[rt-app] <notice> [0] Exiting.

vingu-linaro:
Could you please update tutorial.txt with new properties

src/rt-app.c
vingu-linaro: Can't you get the data directly from data->curr_sched_data ?
When you set only uclamp_{min,max}, we create a sched_data_t with policy == same == -1
this will be replaced by current sched_policy

Also I'm not sure of the behavior at runtime.
It is assumed that if policy == same then we have a priority set correctly in the sched_data which is not the case if only uclamp_min/max has been set

qais-yousef:
> We have a regression with this branch on the json file parsing.
with this branch :
$ rt-app doc/examples/template.json
[rt-app] [json] Error while parsing config
[json]
whereas with master:
$ rt-app doc/examples/template.json
[rt-app] Calibrate ns per loop
[rt-app] pLoad = 19ns : calib_cpu 0
[rt-app] [0] starting thread ...

>[rt-app] [0] Starting with SCHED_OTHER policy with priority 0
[rt-app] [0] Exiting.

Did you try rebasing the branch on top of latest master? I think you're hitting #85

src/rt-app.c
qais-yousef: Hmm I thought @derkling was going to remove the conditional compilation of deadline and always compile it in

vingu-linaro:
No I haven't tried. I have only used the branch of the pull request

src/rt-app.c
qais-yousef: When you say policy == same == -1; do you mean from rt-app point of view or the kernel?

AFAIU the problem is that to do the sched_setattr() syscall, the policy can't be -1.

If you meant we already know it's deadline and we already know the priority so we can fill those directly without doing the extra syscall then yeah. But this is more robust to avoid accidentally changing an attribute that we didn't intend to modify?

qais-yousef:
> Could you please update tutorial.txt with new properties ?

Sure. If this is the only change we end up needing (pending the discussion on the other issues to clear up) - it'd be good to send it as a separate PR unless @derkling has some bandwidth to update this PR himself

qais-yousef:
> No I haven't tried. I have only used the branch of the pull request

Could you try please? Because I can't reproduce on my rebased branch.

By the way - apologies for the delayed response. I was on holiday all of last week.

src/rt-app.c
qais-yousef: > The values in sched_data_t must be the ones set to the task or there is a bug. So it's not more robust to call sched_getattr but it's more time consuming

This doesn't make sense to me. sched_getattr() is the robust way to fill the struct sched_attr. Doing it manually encourages more programming mistakes since what we really care about here is maintain the status queue; bar modify the uclamp values. This is the correct way to do RMW as far as I can tell. Relying on the cached values is error prone since it's not  a single valued structure and could grow with time. So we are adding more complexity and increased probability of breakage - for no real gain. The syscall should be pretty fast and the uclamp values shouldn't be changed a lot except for tests that want to stress uclamp code paths.

vingu-linaro: the uclamp fields have been added in sched_data_t, set_thread_priority and parse_sched_data so they have to reuse the fields which have been already filled with relevant informations.
Then syscall is always more costly than reading the value from a field

qais-yousef: I don't understand the basis of your objection still. How is this breaking anything? I don't see the syscall a performance bottleneck; and the logic is correct as it's doing the RMW in a robust and future proof way.

If the author chose to do it the way you suggest then I'm fine with this. But I don't see a strong technical merit to one way over the other and I think we should respect the author's preference.

qais-yousef: I have the following change; does this help getting over this deadlock?

```
index 040cadf..1780518 100644
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -982,23 +982,6 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
                sa_params.sched_flags = SCHED_FLAG_KEEP_ALL;
                tid = gettid();

-               /*
-                * Currently we cannot use the syscall to set a clamp
-                * value without specifing a valid policy and priotiry.
-                * Read the current values and use them for the following
-                * syscall.
-                * TODO: fix __sched_setscheduler() to get rid of this!
-                */
-               ret = sched_getattr(tid, &sa_params,
-                                   sizeof(struct sched_attr), flags);
-               if (ret != 0) {
-                       log_critical("[%d] sched_getattr returned %d",
-                                       data->ind, ret);
-                       errno = ret;
-                       perror("sched_getattr");
-                       exit(EXIT_FAILURE);
-               }
-
                if (sched_data->util_min != -1) {
                        sa_params.sched_util_min = sched_data->util_min;
                        sa_params.sched_flags |= SCHED_FLAG_UTIL_CLAMP_MIN;
@@ -1018,6 +1001,13 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
                                   sched_data->util_max);
                }

+               /*
+                * we can't change clamp values without valid policy and
+                * prioroity values.
+                */
+               sa_params.sched_policy = sched_data->policy;
+               sa_params.sched_priority = sched_data->prio;
+
                ret = sched_setattr(tid, &sa_params, flags);
                if (ret != 0) {
                        log_critical("[%d] sched_setattr returned %d",
```

qais-yousef: I still have to validate it all works as expected. I prefer to push this as a separate PR on top of this though. I am carrying other changes to remove the --with-deadline option from configure and always build with deadline by default.

qais-yousef: #91 contains all the changes I mentioned. Hopefully this will help to get this merged ASAP. The only issue I can see with this PR as that it can fail to build if building without deadline for the reason mentioned by @credp; if you can edit the file before merging to remove the DLSCHED guards around declaration of sa_params in set_scheduler_priority() that will be great. My PR addresses the issue anyway by removing the optional build of deadline.

vingu-linaro:
This pull request has been superseded by [#91 ](https://github.com/scheduler-tools/rt-app/pull/91)

PR #83, opened 2019-08-15
Fix usage of errno in rt-app.c

qais-yousef:
Some functions return 0 on success and a nonezero error number on
failure. Others return 0 on success and -1 on failure and they set the
errno.

First include <errno.h> to correctly get the definition of errno then
don't set errno manually if the function return -1 and sets the errno
itself.

Signed-off-by: Qais Yousef <qais.yousef@arm.com>

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -36,6 +36,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <errno.h>
 
 #include "config.h"
 #include "rt-app_utils.h"
@@ -50,7 +51,6 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 #define FORKS_LIMIT		1024
 
-static int errno;
 static volatile sig_atomic_t continue_running;
 static pthread_t *threads;
 static int nthreads;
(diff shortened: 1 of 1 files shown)


qais-yousef:
@jlelli @vingu-linaro @credp 

If any of you can spend sometime to review this that'd be appreciated. Thanks!

src/rt-app.c
```diff
@@ -50,7 +51,6 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 #define FORKS_LIMIT		1024
 
-static int errno;
```
credp: This looks like the right change to make - I think you're never supposed to define your own errno, rather include the header.

src/rt-app.c
```diff
@@ -927,7 +927,6 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 				if (ret != 0) {
 					log_critical("[%d] setpriority "
 					     "returned %d", data->ind, ret);
-					errno = ret;
```
credp: I think we should never be setting errno directly when detecting errors from library calls so LGTM.

src/rt-app.c
```diff
@@ -1017,8 +1015,7 @@ void *thread_body(void *arg)
 	{
 		log_notice("[%d] Locking pages in memory", data->ind);
 		ret = mlockall(MCL_CURRENT | MCL_FUTURE);
-		if (ret < 0) {
-			errno = ret;
+		if (ret != 0) {
```
credp: This change is fine, but seems to me a matter of style as this fn can only return 0 or -1. I don't mind the change though, it is clearer that 0 is the only acceptable result.

qais-yousef: I made this change to be consistent with the rest of the code that checks for (ret != 0)

PR #85, opened 2019-09-04
rtapp/config: Fix "ftrace" for backward support

derkling:
The "ftrace" JSON option has always been considered optional and, by default, no
ftrace events was genererated if that option was not specified.

With the addition of the ftrace category support in:

   commit a2d7d8c571eb ("ftrace: Add support for ftrace categories and make them easy to parse")

and specifically with code to grant backward compatibility for the now
deprecated bool type, the JSON parser complains if a value for "ftrace" is not
specified, thus failing to be completely backward compatibile.

Fix that by adding a check on "ftrace" missing and allowing the code to smoothly
fall-back to the default behavior.

Signed-off-by: Patrick Bellasi <patrick.bellasi@arm.com>

### DIFF
--- a/doc/examples/browser-long.json
+++ b/doc/examples/browser-long.json
@@ -123,7 +123,6 @@
 	"global" : {
 		"default_policy" : "SCHED_OTHER",
 		"duration" : 600,
-		"ftrace" : "none",
 		"gnuplot" : false,
 		"logdir" : "./",
 		"log_basename" : "web",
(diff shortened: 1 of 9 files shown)


derkling:
@vingu-linaro this is to fix/complete the backward compatibility you requested in #80.
Unfortunately, despite begin "ftrace" optional we have only examples always specifying it and thus I did not detect the failure when don't have it. I've got it now since in https://github.com/ARM-software/lisa we currently never generate an "ftrace" option.

PR #91, opened 2019-10-25
Add uclamp support improvements

qais-yousef:
Some improvements on top of #81 which addresses some review comments and gets rid of optional building with sched-deadline which is something that creeped up while discussing the PR. It relies on #81 being merged first as it builds on top.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,13 +18,7 @@ AC_CHECK_LIB([rt], [clock_gettime])
 AC_CHECK_LIB([json-c], [json_object_from_file], [], [AC_MSG_ERROR([json-c libraries required])])
 AC_CHECK_FUNCS(sched_setattr, [], [])
 
-AC_ARG_WITH([deadline],
-	[AS_HELP_STRING([--with-deadline], 
-		[Add support for SCHED_DEADLINE])],
-	[AC_DEFINE([DLSCHED], [1], [Define if you have SCHED_DEADLINE support])], 
-	[with_deadline=no])
-
-AM_CONDITIONAL([SET_DLSCHED], [test x$with_deadline != xno && test !HAVE_SCHED_SETATTR])
+AM_CONDITIONAL([SET_DLSCHED], [test !HAVE_SCHED_SETATTR])
 
 AC_ARG_VAR([LOGLVL], [verbosity level, from 0 to 100. 100 is very verbose])
 
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -1008,12 +1008,18 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 			sa_params.sched_flags |= SCHED_FLAG_UTIL_CLAMP_MIN;
 			log_notice("[%d] setting util_min=%d",
 				   data->ind, sched_data->util_min);
+			log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+				   "rtapp_attrs: event=uclamp util_min=%d",
+				   sched_data->util_min);
 		}
 		if (sched_data->util_max != -1) {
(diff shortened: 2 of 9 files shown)


Makefile.am
```diff
@@ -1,6 +1,2 @@
-if SET_DLSCHED
```
vingu-linaro: This can't be removed unconditionally.
We build libdl only if libc doesn't support sched_setattr.

configure.ac
```diff
@@ -18,14 +18,6 @@ AC_CHECK_LIB([rt], [clock_gettime])
 AC_CHECK_LIB([json-c], [json_object_from_file], [], [AC_MSG_ERROR([json-c libraries required])])
 AC_CHECK_FUNCS(sched_setattr, [], [])
 
-AC_ARG_WITH([deadline],
-	[AS_HELP_STRING([--with-deadline], 
-		[Add support for SCHED_DEADLINE])],
-	[AC_DEFINE([DLSCHED], [1], [Define if you have SCHED_DEADLINE support])], 
-	[with_deadline=no])
-
-AM_CONDITIONAL([SET_DLSCHED], [test x$with_deadline != xno && test !HAVE_SCHED_SETATTR])
```
vingu-linaro: this should be kept

src/rt-app.c
```diff
@@ -1018,6 +1001,13 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 				   sched_data->util_max);
 		}
 
+		/*
+		 * we can't change clamp values without valid policy and
+		 * prioroity values.
```
vingu-linaro: s/prioroity/priority/

src/rt-app.c
```diff
@@ -1018,6 +1001,13 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 				   sched_data->util_max);
 		}
 
+		/*
+		 * we can't change clamp values without valid policy and
```
vingu-linaro: uppercase: We

Makefile.am
qais-yousef: I remember in our call when Patrick was here we agreed to always compile deadline in as it's been around for long enough now. Am I mis-remembering?

src/rt-app.c
qais-yousef: done

src/rt-app.c
qais-yousef: done

Makefile.am
vingu-linaro: Yes it's right, you don't mis-remember but SET_DLSCHED is not about compiling the support of deadline scheduler but about compiling a local support of sched_setattr if the lib doesn't support it.
So DLSCHED must be removed but not SET_DLSCHED

qais-yousef: Ah I see. Okay let me see if I can hack something and updated the PR

vingu-linaro: It should be just keeping this : if SET_DLSCHED in Makefile.am
as well as the AM_CONDITIONAL([SET_DLSCHED] .... in configure.ac

qais-yousef: Okay thanks. Done that. Just to confirm, all the `#ifdef DLSCHED` in the *.[ch] files should be removed now, right?

qais-yousef: And the configure.ac test is now shorter

vingu-linaro: > Okay thanks. Done that. Just to confirm, all the `#ifdef DLSCHED` in the *.[ch] files should be removed now, right?
Yes.

qais-yousef:
Thanks for the review @vingu-linaro. If there are no more reviewers intending to look at this it'd be nice if we can merge #81 then this one.

PR #92, opened 2019-10-29
barrier: fix waiting event with multi instances tasks

vingu-linaro:
The events of  thetask object are parsed only once even if several
instances will be then created.
barrier.waiting must add all instances of the task that will sync.
Similar init can't be done for fork event has we don't know in advance
how many forkees will be created which means that fork and barrier can't
be used in the same task description. The solution would be to set an init
value to the barrier resource in the "resources" object but that's a
dedicated dev.


vingu-linaro:
I have added a comment in the tutorial which is probably enough for now. user also have to understand what they can do and what they can't do.
Then, setting the barrier with a default value in the resources object could be a solution to make barrier and fork work together

PR #94, opened 2020-01-17
rt-app:config: remove spurious log_notice

vingu-linaro:
rt-app displays the spurious and useless :
  [rt-app] <notice> [json]         util_min: -1, util_max: -1
for each and every threads node that is parsed even if no clamp values
have been specified in the json node.

Remove this useless print


vingu-linaro:
@jlelli @credp 
This is just a simple cleanup of the log which is quite painful and appears for each and every exec of rt-app

PR #95, opened 2020-01-17
Initial numa support

PeterPuhov-Linaro:
This PR will provide basic control for task NUMA memory allocation policy.
It will help to test CONFIG_NUMA_BALANCING logic and it's interaction with CFS scheduler.
This code is integrated and tested with LISA on both arm64 and x86_64 target architectures.

### DIFF
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -282,6 +282,47 @@ util_max: Integer. Can be specified at task level or phase level. Sets the max
 utilization value the task can have; which helps capping the task from
 scheduler point of view.
 
+*** NUMA memory binding
+
+* membind_nodes: Array of Integer. Define the NUMA binding of the thread.
+Default value is all NUMA nodes of the system. 
+An example : "membind_nodes" : [0, 2, 3]
+
+"membind_nodes" Can be specified at task level or phase level. 
+As an example, below creates two threads.
+One thread will run with memory binding to nodes 2 and 3. 
+The second task will run first phase with memory binding to nodes 0 and 1, 
+second phase with memory binding to node 2.
+Please note, that we follow an "event based policy", which means that 
+rt-app changes memory binding when there is a "membind_nodes" event 
+and don't do anything otherwise.
+
+"tasks" : {
+    "thread1" : {
+        "membind_nodes" : [2,3],
+        "phases" : {
+            "phase1" : {
+                "run" : 10,
+                "sleep" : 10
+            }
+        }
+    }
+    "thread2" : {
+        "phases" : {
+            "phase1" : {
+                "membind_nodes" : [0,1],
+                "run" : 10,
+                "sleep" : 10
+            }
+            "phase2" : {
+                "membind_nodes" : [2],
+                "run" : 10,
+                "sleep" : 10
+            }
+        }
+    }
+}
+
 *** taskgroups
 
 * taskgroup: String. Can be specified at task level or phase level. Currently
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -849,6 +849,27 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
 	}
 }
 
+static void set_thread_membind(thread_data_t *data, numaset_data_t * numa_data)
+{
+#if HAVE_LIBNUMA
+	if (numa_data->numaset == NULL)
+		return;
+
+	if(data->curr_numa_data == numa_data)
+		return;
+
+	if (data->curr_numa_data == NULL ||
+			!numa_bitmask_equal(numa_data->numaset, data->curr_numa_data->numaset))
+	{
+		log_debug("[%d] setting numa_membind to Node (s) %s", data->ind,
+				numa_data->numaset_str);
+		numa_set_membind(numa_data->numaset);
+	}
+	data->curr_numa_data = numa_data;
+#endif
+}
+
+
 static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
 {
 	struct sched_param param;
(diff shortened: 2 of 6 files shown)


vingu-linaro:
Hi Peter,

Thanks for proposing to add numactl into rt-app.
I have started to review your pull request but I can't see a description of the new "node" event in tutorial.txt. Could you add a description of this "node" in tutorial.txt ?
 Also you seem to apply this only at the start of the threads but would it make sense to be able to change the node affinity at runtime like cpu affinity, scheduling policy and uclamp setting ?

vingu-linaro:
> Sure, I will add description of the new "node" event to tutorial.txt
> I can also allow to change node affinity at runtime.
> I think it is not common use case. Typical HPC workloads select a set of nodes
> (sharing same PCI resources or HW accelerators) and set NUMA affinity to make sure that memory
> will be allocated from proper nodes. Changing that affinity at runtime may result in performance degradation since some memory is already allocated from other nodes and kernel NUMA_BALANCING algorithm may not work properly.

I agree that's the most common case. My goal is mainly to keep the same behavior for all properties like cpu affinity, scheduling policy etc... and the numa one

> However, if you think that it will be more consistent to allow node affinity changes at runtime
> like cpu affinity, etc. I will definitely make this change.
> Please advise.

doc/tutorial.txt
```diff
@@ -272,6 +272,48 @@ inherit the affinity from the task level):
...
+
+* nodes: Array of Integer. Define the NUMA binding of the thread. 
+Default value is all NUMA nodes of the system. An example :  "nodes" : [0, 2, 3]
+
+"nodes" can be specified at task level or phase level. As an example, below
+creates two threads. One thread will run with memory binding to nodes 2 and 3. The
+second task will run first phase with memory binding to nodes 0 and 1, 
+second phase with memory binding to node 2, 
+and the third phase with memory binding to nodes 4, 5, and 6 (it will inherit memory binding from the task level):
+
```
vingu-linaro: I would prefer that you don't change the memory binding for the 3rd phase to be aligned with "policy", "priority", cgroups and utilization clamping instead of cpuset. I prefer to follow an event based policy than an inheritance which means that rt-app changes memory binding when there is a "nodes" event and don't do anything otherwise. "cpuset" is prone to mistake as it can change the affinity even if you haven't put anything in a phase.

This behavior will also simplify quite a lot the code as you don't need to create new string for default node affinity

src/rt-app.c
```diff
@@ -849,6 +849,108 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
 	}
 }
 
+#if HAVE_LIBNUMA
+static int create_numaset_str(numaset_data_t *numa_data)
+{
```
vingu-linaro: witha event based policy you don't need this anymore

src/rt-app.c
```diff
@@ -849,6 +849,108 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
...
+}
+#endif
+
+static void set_thread_membind(thread_data_t *data, numaset_data_t * numa_data)
+{
+#if HAVE_LIBNUMA
+	numaset_data_t *actual_numa_data = &data->numa_data;
+
+	if (data->def_numa_data.numaset == NULL) {
+		/* Get default numa bind mask */
```
vingu-linaro: this def_numa_data is no more needed

doc/tutorial.txt
PeterPuhov-Linaro: Sure , will do.

src/rt-app.c
PeterPuhov-Linaro: Will remove.

src/rt-app.c
PeterPuhov-Linaro: Will remove

src/rt-app.c
```diff
@@ -1133,6 +1156,7 @@ void *thread_body(void *arg)
 		struct timespec t_diff, t_rel_start;
 
 		set_thread_affinity(data, &pdata->cpu_data);
+		set_thread_membind(data, &pdata->numa_data);
```
vingu-linaro: minor: Could you keep the same ordering  here and above, ie:
set_thread_priority
set_thread_membind
set_thread_taskgroup

vingu-linaro:
@jlelli , @credp 
Do you have comment on this pull request ?

doc/tutorial.txt
```diff
@@ -282,6 +282,45 @@ util_max: Integer. Can be specified at task level or phase level. Sets the max
 utilization value the task can have; which helps capping the task from
 scheduler point of view.
 
+*** NUMA memory binding
+
+* nodes: Array of Integer. Define the NUMA binding of the thread. 
+Default value is all NUMA nodes of the system. An example : "nodes" : [0, 2, 3]
+
+"nodes" Can be specified at task level or phase level. As an example, below
```
jlelli: This only affects memory, right?
Wonder if we'd better add this information to the keyword we use, so to be able to differentiate between, say, membind and bind (which affects affinity as well).

src/rt-app.c
PeterPuhov-Linaro: Sure, will do.

src/rt-app_parse_config.c
```diff
@@ -785,6 +785,48 @@ static void parse_cpuset_data(struct json_object *obj, cpuset_data_t *data)
...
+		unsigned int i;
+		unsigned int node_idx;
+		unsigned int max_node;
+
+#if HAVE_LIBNUMA
+		if(numa_available() == -1)
+#endif
+		{
+			log_critical(PIN2 "NUMA is not available");
+			exit(EXIT_INV_CONFIG);
```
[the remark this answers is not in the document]
PeterPuhov-Linaro: Will do

doc/tutorial.txt
PeterPuhov-Linaro: I can change it to something like "membind_nodes"
Is it ok?

PR #97, opened 2020-02-21
Added strict event length match

PeterPuhov-Linaro:
This fix will guarantee exact match of "special" keys such as:
"lock", "unlock", "wait", "signal", etc.
This will help to prevent accidental matches like:
"mem" and "membind_nodes"

### DIFF
--- a/src/rt-app_parse_config.c
+++ b/src/rt-app_parse_config.c
@@ -743,7 +743,8 @@ obj_is_event(char *name)
 
     for (pos = events; *pos; pos++) {
 	    char *event = *pos;
-	    if (!strncmp(name, event, strlen(event)))
+	    if (strlen(event) == strlen(name) &&
+	    		!strncmp(name, event, strlen(event)))
 		    return 1;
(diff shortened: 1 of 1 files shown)


vingu-linaro:
we can't make strict match because each json object must have a unique id  so if we need to use several time the "run" event in the same phase as an example, then we will use "run1" "run2" "run3" ... In rder to make the json file easier to write and read, you don't have to bother with this and can always use the "run" name and then use workgen to generate a file with unique id for each event.
That being said, we completely missed the situation that membind_nodes could be interpreted as a mem event. So either we rename membind_nodes to not match an event or we add an exception in function that looks for events to not consider an id starting with "membind_nodes" as a "mem" event.
TBH the 1st solution seems to be the easiest one

PeterPuhov-Linaro:
I am planning to rename "membind_nodes" to "nodes_membind"
Is it ok?

vingu-linaro:
nodes_membind looks good to me

PR #99, opened 2020-05-21
rt-app: Fix sched_setattr() failures with older kernels

psk243:
When running rt-app on older android kernels (4.19 and before),
sched_setattr() syscall fails with -E2BIG error code. These kernels
don't have uclamp support. The kernel returns an error when the
new fields (in this case uclamp attributes) not known to kernel
are non zero. To fix this initialize the sched_attr structure to 0.

Signed-off-by: Pavankumar Kondeti <pkondeti@codeaurora.org>


vingu-linaro:
looks good to me

PR #100, opened 2020-06-17
Fix gnuplot logs file name for single thread task

thymbahutymba:
Wrong logs file when using gnuplot. Taking as reference the following config file:
```
{
  "tasks": {
    "task3": {
      "runtime": 3000,
      "timer": {
        "period": 1000000,
        "mode": "absolute",
        "ref": "unique"
      }
    },
    "task1": {
      "runtime": 19000,
      "timer": {
        "period": 1000000,
        "mode": "absolute",
        "ref": "unique"
      }
    }
  },
  "global": {
    "duration": 2,
    "default_policy": "SCHED_DEADLINE",
    "pi_enabled": false,
    "lock_pages": false,
    "logdir": "./rtapp-log",
    "log_basename": "rt-app",
    "log_mode": "file",
    "gnuplot": true
  }
}
```
The names of the log files are: `rt-app-task1-1.log  rt-app-task3-0.log`, but taking a look at the `rt-app-period.plot` file we have: 
```
plot "rt-app-task3-0-0.log" u ($5/1000):4 w l title "thread [task3-0] (SCHED_DEADLINE)", "rt-app-task1-1-1.log" u ($5/1000):4 w l title "thread [task1-1] (SCHED_DEADLINE)"
```

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -1401,22 +1401,12 @@ static void setup_main_gnuplot(void)
 			"plot ", tmp);
 
 		for (i=0; i<running_threads; i++) {
-			if (opts.nthreads == opts.num_tasks) {
-				fprintf(gnuplot_script,
-					"\"%s-%s.log\" u ($5/1000):4 w l"
-					" title \"thread [%s] (%s)\"",
-					opts.logbasename, threads[i].data->name,
-					threads[i].data->name,
-					policy_to_string(threads[i].data->sched_data->policy));
-			} else {
-				fprintf(gnuplot_script,
-					"\"%s-%s-%d.log\" u ($5/1000):4 w l"
-					" title \"thread [%s] (%s)\"",
-					opts.logbasename, threads[i].data->name,
-					threads[i].data->ind,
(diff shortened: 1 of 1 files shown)


src/rt-app.c
```diff
@@ -1401,13 +1401,22 @@ static void setup_main_gnuplot(void)
...
 
 		for (i=0; i<running_threads; i++) {
-			fprintf(gnuplot_script,
-				"\"%s-%s-%d.log\" u ($5/1000):4 w l"
-				" title \"thread [%s] (%s)\"",
-				opts.logbasename, threads[i].data->name,
-				threads[i].data->ind,
-				threads[i].data->name,
-				policy_to_string(threads[i].data->sched_data->policy));
+			if (opts.nthreads == opts.num_tasks) {
```
vingu-linaro, line 1404: There is no need to make a difference between these 2 cases because threads[i].data->ind is not needed in every case

thymbahutymba: Ok, I thought they were two different cases.

Issue #38, opened 2017-08-30
set affinity for a thread scheduled by shed-deadline


zl19941027:
Since affinity setting is currently disabled for tasks handled by sched-deadline, how can we restrict threads to certain cpus but using the policy sched-deadline ?

zl19941027:
Is it possible to reset the "cpumask_t" and "cpus_allowed" for every thread ( in the code of "cpus" ) ?

vingu-linaro:
hi z-linlinlin,

As you mentioned above, we can't restrict cpus allowed for deadline scheduler so it's up to the writer of the json file use case to ensure that this will not happen.
as default cpumask is all cpus and we can set affinity per phase with a fall back to this default mask when not mentioned, you should be able to ensure that all cpus are allowed when setting deadline scheduling mode

credp:
IIUC, deadline performs the admittance test only once when a task is changed to deadline class, which means we can't be changing affinity after thread creation. However I think it should be possible to use cpusets to restrict deadline tasks so long as you don't go changing them after creation - but you would need to add better cgroup support to rt-app first. See https://github.com/scheduler-tools/rt-app/pull/49



From the deadline docs in-kernel:

5. Tasks CPU affinity
=====================

 -deadline tasks cannot have an affinity mask smaller that the entire
 root_domain they are created on. However, affinities can be specified
 through the cpuset facility (Documentation/cgroup-v1/cpusets.txt).

5.1 SCHED_DEADLINE and cpusets HOWTO
------------------------------------

 An example of a simple configuration (pin a -deadline task to CPU0)
 follows (rt-app is used to create a -deadline task).

 mkdir /dev/cpuset
 mount -t cgroup -o cpuset cpuset /dev/cpuset
 cd /dev/cpuset
 mkdir cpu0
 echo 0 > cpu0/cpuset.cpus
 echo 0 > cpu0/cpuset.mems
 echo 1 > cpuset.cpu_exclusive
 echo 0 > cpuset.sched_load_balance
 echo 1 > cpu0/cpuset.cpu_exclusive
 echo 1 > cpu0/cpuset.mem_exclusive
 echo $$ > cpu0/tasks
 rt-app -t 100000:10000:d:0 -D5 (it is now actually superfluous to specify
 task affinity)

jlelli:
> performs the admittance test only once when a task is changed to deadline class

[...]

PR #101, opened 2020-06-22
Fix how period is taken from timer

thymbahutymba:
The period was not correctly extracted from the timer of the task. The period was extracted from the task object, which has no period object (according to the doc). Therefore the period was the same as the runtime which leads to an error using SCHED_DEADLINE. This problem has been exposed in https://github.com/scheduler-tools/rt-app/issues/38#issuecomment-646639431. In order to understand the problem let's consider the following config file:
```
{
  "tasks": {
    "task3": {
      "runtime": 233,
      "timer": {
        "period": 1000,
        "mode": "absolute",
        "ref": "unique"
      }
    },
    "task4": {
      "runtime": 115,
      "timer": {
        "period": 1000,
        "mode": "absolute",
        "ref": "unique"
      }
    },
    "task0": {
      "runtime": 123,
      "timer": {
        "period": 1000,
        "mode": "absolute",
        "ref": "unique"
      }
    },
    "task2": {
      "runtime": 128,
      "timer": {
        "period": 1000,
        "mode": "absolute",
        "ref": "unique"
      }
    },
    "task1": {
      "runtime": 151,
      "timer": {
        "period": 1000,
        "mode": "absolute",
        "ref": "unique"
      }
    }
  },
  "global": {
    "duration": 2,
    "default_policy": "SCHED_DEADLINE",
    "pi_enabled": false,
    "lock_pages": false,
    "logdir": "./rtapp-log",
    "log_basename": "rt-app",
    "log_mode": "file",
    "gnuplot": false
  }
}
```
Since the period is inside the timer and both `dl-runtime` and `dl-period` were not set, in the "old" code the period is equal to the task runtime which end up in utilization equal 1 for each task. In the issue I said that the config worked only  with N_TASK <= NCPUS - 1 because, if we consider the default parameter of `sched_rt_period_us` (1000000) and `sched_rt_runtime_us` (950000), for two cpus we have that the max utilization allowed by SCHED_DEADLINE is `2 * sched_rt_runtime_us / sched_rt_period_us` that is equal to 1,9. Since each task has U = 1 the config worked, in this case, with only one task. 
I'm not sure if that was desired behaviour, maybe the point was to use `dl-runtime` and `dl-period` but with the default config it does not work properly.

### DIFF
--- a/src/rt-app_parse_config.c
+++ b/src/rt-app_parse_config.c
@@ -881,8 +881,17 @@ static sched_data_t *parse_sched_data(struct json_object *obj, int def_policy)
 		/* Support legacy grammar for thread object */
 		if (!tmp_data.runtime)
 			tmp_data.runtime = get_int_value_from(obj, "runtime", TRUE, 0);
-		if (!tmp_data.period)
-			tmp_data.period = get_int_value_from(obj, "period", TRUE, tmp_data.runtime);
+
+		if (!tmp_data.period) {
+			struct lh_entry *entry; char *key; struct json_object *val; int idx;
+			foreach(obj, entry, key, val, idx) {
+				if (!strncmp(key, "timer", strlen("timer"))) {
+					tmp_data.period = get_int_value_from(val, "period", TRUE, tmp_data.runtime);
+					break;
+				}
+			}
+		}
+		
 		if (!tmp_data.deadline)
 			tmp_data.deadline = get_int_value_from(obj, "deadline", TRUE, tmp_data.period);
 	}


src/rt-app_parse_config.c
```diff
@@ -881,8 +881,17 @@ static sched_data_t *parse_sched_data(struct json_object *obj, int def_policy)
 		/* Support legacy grammar for thread object */
 		if (!tmp_data.runtime)
 			tmp_data.runtime = get_int_value_from(obj, "runtime", TRUE, 0);
-		if (!tmp_data.period)
-			tmp_data.period = get_int_value_from(obj, "period", TRUE, tmp_data.runtime);
+
+		if (!tmp_data.period) {
```
vingu-linaro, line 885: This is only used to support legacy/old json file when the deadline parameters were define with the key "runtime", "deadline" and "period" and the "runtime" event didn't exist. The set  of deadline parameter is now done with dl-runtime, dl-deadline and dl-period and you must set them.
Your proposal tries to extend this backward compatibility to use the runtime and period key of some events but we will not do so.
a json file which wants to use deadline class must set the key dl-runtime, dl-deadline and dl-period. The only exception is for old json file that were using the old keys: runtime, deadline and period but not to try to estimate the value based on some events

thymbahutymba: So in order to define a deadline task I need both (runtime, timer->period) and (dl-runtime, dl-period)?. For example:
```
{
  "tasks": {
    "task1": {
      "runtime": 30000,
      "timer": {
        "period": 65000,
        "mode": "absolute",
        "ref": "unique"
      },
      "dl-runtime": 30000,
      "dl-period": 65000
    },
    [...]
}
```
With maybe runtime = 0.95 * dl-runtime?

vingu-linaro: Yes. you must set at least dl-runtime and dl-period. Then dl_deadline = dl_period if not set

thymbahutymba: Perfect, then in order to avoid that the task misses its deadline I should use runtime < dl-runtime? let's say: runtime = 95% of dl-runtime.

One more request: could you please add at least one example that uses SCHED_DEADLINE with these parameters? It will be very helpful for who come next :)

vingu-linaro: The ratio runtime vs dl-runtime is a bit platform and use case specific because it will depends if the runtime is in one run or split in several small ones as an example but 95% is probably a sane value. @jlelli  do you have any advice for this ratio ?
Please provide an example with deadline parameters that makes sense to you and I will merge it

thymbahutymba: I prepared an example with utilization equal to 1.5 considering runtime = 95% of dl-runtime.
```
{
  "tasks": {
    "task2": {
      "runtime": 12059,
      "timer": {
        "period": 25000,
        "mode": "absolute",
        "ref": "unique"
      },
      "dl-runtime": 12694,
      "dl-period": 25000
    },
    "task1": {
      "runtime": 11108,
      "timer": {
        "period": 15000,
        "mode": "absolute",
        "ref": "unique"
      },
      "dl-runtime": 11693,
      "dl-period": 15000
    },
    "task3": {
      "runtime": 1863,
      "timer": {
        "period": 50000,
        "mode": "absolute",
        "ref": "unique"
      },
      "dl-runtime": 1962,
      "dl-period": 50000
    },
    "task4": {
      "runtime": 6970,
      "timer": {
        "period": 55000,
        "mode": "absolute",
        "ref": "unique"
      },
      "dl-runtime": 7337,
      "dl-period": 55000
    },
    "task0": {
      "runtime": 380,
      "timer": {
        "period": 10000,
        "mode": "absolute",
        "ref": "unique"
      },
      "dl-runtime": 400,
      "dl-period": 10000
    }
  },
  "global": {
    "duration": 4,
    "default_policy": "SCHED_DEADLINE",
    "pi_enabled": false,
    "lock_pages": false,
    "logdir": "./rtapp-log",
    "log_basename": "rt-app",
    "log_mode": "file",
    "gnuplot": true
  }
}
```

jlelli: > The ratio runtime vs dl-runtime is a bit platform and use case specific because it will depends if the runtime is in one run or split in several small ones as an example but 95% is probably a sane value. @jlelli do you have any advice for this ratio ?

Not much to add actually. Starting from 95% of dl-runtime seems a pretty good rule of thumb.

PR #105, opened 2020-09-15
Fix unchanged prio

vdonnefort:


src/rt-app.c
```diff
@@ -1031,6 +1031,23 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
...
+		ret = sched_getattr(tid, &sa_params,
+				    sizeof(struct sched_attr), flags);
+		if (ret != 0) {
+			log_critical("[%d] sched_getattr returned %d",
+					data->ind, ret);
+			errno = ret;
+			perror("sched_getattr");
+			exit(EXIT_FAILURE);
+		}
+
```
vingu-linaro, line 1031: I remember discussing this with Qais.
We already know which policy is currently set so we don't need the syscall which is costly

src/rt-app.c
```diff
@@ -989,93 +978,135 @@ static void set_thread_priority(thread_data_t *data, sched_data_t *sched_data)
...
+			break;
+		default:
+			log_error("Unknown scheduling policy %d",
+				  sched_data->policy);
 			exit(EXIT_FAILURE);
-		}
 	}
 
+	_set_thread_priority(data, sched_data);
+	_set_thread_deadline(data, sched_data);
```
vingu-linaro, line 1107: deadline should be part of _set_thread_priority. There is no reason to make it special

src/rt-app.c
vdonnefort: The problem isn't the policy but the priority. In RT case, `sched_data->prio` is `sched_priority`, it is working. In the case of CFS though, `sched_priority` has no meaning and it should be set to a default value (while `sched_data->prio` is the nice value (which can be negative... while `sched_priority` is unsigned). We could hardcode this value here, but there's no guarantee that it won't change in the kernel. That's why I believe it is better to read the `sched_priority` value, instead of writing it.

In any case, `sa_params.sched_priority = sched_data->prio` is wrong. See `__sched_setscheduler`

```        /*
         * Valid priorities for SCHED_FIFO and SCHED_RR are
         * 1..MAX_USER_RT_PRIO-1, valid priority for SCHED_NORMAL,
         * SCHED_BATCH and SCHED_IDLE is 0.
         */
```

src/rt-app.c
vdonnefort: It was to avoid having something weird in `_set_thread_priority()`: 

```
/*
 * If THREAD_PRIORITY_UNCHANGED, but we are deadline, it's probably because we
 * want to edit one of the deadline parameters.
 */
if (sched_data->prio == THREAD_PRIORITY_UNCHANGED &&
     sched_data->policy != deadline)
          return;
```

But I can make the change if you'd rather prefer to have all the policies covered in _set_thread_priority.

src/rt-app.c
vingu-linaro: so set sa_params.sched_priority to 0 for CFS policy

src/rt-app.c
vingu-linaro: The rule is  that : if there is a sched_data, it's means that there is something to change on the scheduler policy/prio. Now,

src/rt-app.c
```diff
@@ -913,174 +913,185 @@ static int __sched_priority(sched_data_t *sched_data)
...
+					sched_data_t *sched_data)
 {
 	struct sched_param param;
-	policy_t policy;
+	int ret;
+
+	/* Priority unchanged => Policy unchanged */
+	if (sched_data->prio == THREAD_PRIORITY_UNCHANGED)
+		return;
+
```
vingu-linaro, line 925: this is not enough, we can set new priority without changing policy. In the cfs case, we don't want to call pthread_setschedparam

vingu-linaro (changes requested):

src/rt-app.c
```diff
@@ -913,174 +921,203 @@ static int __sched_priority(sched_data_t *sched_data)
...
+	/*
+	 * In the CFS case, sched_data->prio is the NICE value. As long as the
+	 * policy hasn't changed, there's no need to call
+	 * __set_thread_policy_priority()
+	 */
+	if (sched_data->policy == same)
```
vingu-linaro, line 989: if (sched_data->policy != same) ?

vdonnefort: Woops :( Will fix that in the next version

src/rt-app.c
```diff
@@ -897,175 +897,227 @@ static void set_thread_membind(thread_data_t *data, numaset_data_t * numa_data)
...
-		ret = sched_setattr(tid, &sa_params, flags);
-		if (ret != 0) {
-			log_critical("[%d] sched_setattr returned %d",
-				     data->ind, ret);
-			perror("sched_setattr: failed to set uclamp value(s)");
-			exit(EXIT_FAILURE);
-		}
-	}
-
+	sched_data->policy = __sched_policy(data, sched_data);
```
vdonnefort, line 1120: The inconvenience with this approach, compared to bringing a new `sched_data` member for holding "same" , is that the `sched_data` is modified.

In the case of rt-app, looping over few phases, the first iteration will overwrite the `sched_data->policy`, and we'll loose the information "same" for the second iteration.

Is it still ok to pursue with this approach ?

vdonnefort: I could alternatively bring a new member into `thread_data_t` instead of `sched_data_t` ?

vingu-linaro: Setting sched_data->policy with policy of the curr_sched_data->policy if current value is same is the current behavior and works just fine. Then compare sched_data->policy with curr_sched_data->policy to know if you need to change the policy instead of if (sched_data->policy != same) which will work only for the 1st time we run the phase

vdonnefort: The problem I have is with the following setup:

```
phase_0 {
    "policy": SCHED_OTHER
}
phase_1 {
    "priority": -10
}
```

```
# First loop
## Phase 0

$ setting scheduler SCHED_OTHER priority 0
# call pthread_setschedparam
# call setpriority

## Phase 1
$ setting scheduler SCHED_OTHER priority -10
# sched_data->policy == same .. skip pthread_setschedparam
# call setpriority to set -10.
# sched_data->policy = SCHED_OTHER

# Second loop

## Phase 0
...

## Phase 1
$ setting scheduler SCHED_OTHER priority -10
# call pthread_setschedparam. ... as now sched_data->policy == SCHED_OTHER
# call setpriority
```

src/rt-app.c
```diff
@@ -913,174 +921,210 @@ static int __sched_priority(sched_data_t *sched_data)
...
-			log_critical("[%d] sched_setattr returned %d",
-				     data->ind, ret);
-			perror("sched_setattr: failed to set uclamp value(s)");
+				  sched_data->policy);
 			exit(EXIT_FAILURE);
-		}
 	}
 
+	/* Assign a policy, != same, for the current sched_data */
+	sched_data->policy = __sched_policy(data, sched_data);
```
vingu-linaro, line 1127: If you put this earlier in the function set_thread_param() (ie before switch (__sched_policy)), you can replace all other calls to __sched_policy() by using directly sched_data->policy because sched_data->policy will always be a valid value.
That's the last remaining point to be fix.
The code is easier to read with your changes

vdonnefort: Indeed, I can remove that portion now.

vingu-linaro (approved):

PR #107, opened 2021-01-04
Taskgroups and deadline fixes

deggeman:


src/rt-app_parse_config.c
```diff
@@ -939,8 +939,10 @@ static taskgroup_data_t *parse_taskgroup_data(struct json_object *obj)
 
 static void check_taskgroup_policy_dep(phase_data_t *pdata, thread_data_t *tdata)
 {
-	/* Save sched_data as thread's current sched_data. */
-	if (pdata->sched_data)
+	/* Save sched_data as thread's current sched_data in case its policy_t is
```
vingu-linaro, line 953: comment style problem
/*
 * text
 */

src/rt-app.c
```diff
@@ -1093,7 +1093,7 @@ static void set_thread_param(thread_data_t *data, sched_data_t *sched_data)
 		return;
 
 	/* if no policy is specified, reuse the previous one */
-	if ((sched_data->policy == same) && data->curr_sched_data)
+	if ((sched_data->policy == none) && data->curr_sched_data)
```
vingu-linaro, line 1096: replacing the -1 value by the enum one makes sense but I really prefer keeping same instead of none because that really what it means same as previous phase

vingu-linaro:
Apart the 2 comments above, everything else looks fine

src/rt-app_parse_config.c
deggeman: Will fix it.

src/rt-app.c
deggeman: OK, will change it.

PR #108, opened 2021-02-04
config: Add "events" key in parse_task_phase_data()

douglas-raillard-arm:
Allow parsing an "events" key for phases with the following format:
```
    "events": [
	{"run": 1000},
	{"timer": {"ref": "helloworld", "period": 16000}},
	{"run": 9000},
	{"sleep": 1000}
    ]
```

This allows:
* Enforced ordering of events. Relying on the order of keys in an
  object is issue-prone, as interfaces to interact with mappings in
  most languages are not designed around preserving such order. On
  the contrary, array operations have always a well defined
  order semantic.

* More critically, this  allows repeating the same event an
  arbitrary number of times. Complex workload might otherwise need
  to be split among different phases to workaround that issue, which
  comes with some overhead and extra bookeeping in external code.

Checked constraints:
* only event objects are allowed in the 
* objects can only have one 
* when "events" is in use, no event key is allowed in the phase
   body. This means the user either uses the old format or the new
   one but mixing is not allowed.

Signed-off-by: Douglas RAILLARD <douglas.raillard@arm.com>


doc/tutorial.txt
```diff
@@ -650,7 +674,7 @@ The generated events are of these main categories:
 
   - rtapp_loop: event=start thread_loop=0 phase=0 phase_loop=0
   - rtapp_loop: event=end thread_loop=0 phase=0 phase_loop=0
-  
+
```
vingu-linaro, line 677: please don't mix typo and new feature in the same patch or PR

doc/tutorial.txt
```diff
@@ -756,5 +780,3 @@ below)
    20000 ++--+----+---+----+---+---+----+---+----+--++ 494000
        17560556057560556057560656065606756065606756075607
                     Loop start time [msec]
-
```
vingu-linaro, line 759: ditto

doc/tutorial.txt
douglas-raillard-arm: Ah looks like my text editor stripped the lines, sorry about that

doc/tutorial.txt
```diff
@@ -339,8 +339,32 @@ unsigned int max_nbr_tgs = 32' at compile time.
...
+    "events": [
+        {"run": 1000},
+        {"timer": {"ref": "helloworld", "period": 16000}},
+        {"run": 9000},
+        {"sleep": 1000}
+    ]
+
+  Each object in the array must contain a single event key. This format allows
+  specifying the same event more than once (like "run" in the example), and is
+  friendlier to languages with map types that are not preserving insertion
```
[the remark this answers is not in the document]
douglas-raillard-arm: Do you mean via workgen or the way rt-app handles things currently ?

src/rt-app_parse_config.c
```diff
@@ -983,37 +983,94 @@ parse_task_phase_data(struct json_object *obj,
...
-	int i;
+	struct json_object *array_val;
+	size_t i;
+	struct json_object *events_array = NULL;
+	bool is_first_entry;
 
 	log_info(PFX "Parsing phase");
 
+	data->nbevents = 0;
+	data->events = NULL;
```
vingu-linaro, line 994: Please try to not move code unnecessarily

src/rt-app_parse_config.c
```diff
@@ -983,37 +983,94 @@ parse_task_phase_data(struct json_object *obj,
...
+			}
+			events_array = val;
+			break;
+		}
+	}
+
+	foreach(obj, entry, key, val, idx) {
+		if (obj_is_event(key)) {
+			/* Check that we are not trying to mix old and new styles */
+			if (events_array) {
```
vingu-linaro, line 1013: if an events array is detected we should simply ignore old style

src/rt-app_parse_config.c
```diff
@@ -983,37 +983,94 @@ parse_task_phase_data(struct json_object *obj,
...
-	foreach(obj, entry, key, val, idx) {
-		if (obj_is_event(key)) {
-			log_info(PIN "Parsing event %s", key);
-			parse_task_event_data(key, val, &data->events[i], tdata, opts);
-			i++;
+	if (events_array) {
+		for (i=0; i < data->nbevents; i++) {
+			array_val = json_object_array_get_idx(events_array, i);
+			is_first_entry = true;
+			foreach(array_val, entry, key, val, idx) {
```
vingu-linaro, line 1043: what is the purpose of this foreach ?
isn't all object inside the array an event

src/rt-app_parse_config.c
```diff
@@ -983,37 +983,94 @@ parse_task_phase_data(struct json_object *obj,
...
+	data->events = NULL;
+
 	/* loop */
 	data->loop = get_int_value_from(obj, "loop", TRUE, 1);
 
-	/* Count number of events */
-	data->nbevents = 0;
 	foreach(obj, entry, key, val, idx) {
-		if (obj_is_event(key))
+		if(!strncmp("events", key, 6)) {
```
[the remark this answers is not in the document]
douglas-raillard-arm: Indeed, I was looking for a function like that in json-c but it turns out it's a helper in that file.

src/rt-app_parse_config.c
douglas-raillard-arm: Is it really an improvement though ? The foreach is still needed to count the number of events anyway and silently ignoring events is not really something users would thank us for.

src/rt-app_parse_config.c
douglas-raillard-arm: The array items are themselves objects:
```
[{event1: param1}, {event2: param2}, ...]
```

The alternative would be to use sub-arrays like a tuple instead of objects, but the object approach is the standard way of doing things, as it is pretty clean in YAML:

```
events:
    - run: 42
    - sleep: 53
 ```

(and the sub-array solution does not save any typing anyway)

vingu-linaro: but that was already the case with the old style and obj_is_event() is there to check that those objects are events

src/rt-app_parse_config.c
vingu-linaro: but when we have found an array we can skip this counting entirely because we don't care about old style events

src/rt-app_parse_config.c
douglas-raillard-arm: After checking parse_task_event_data() it looks like it's doing a validation on the event name and will exit if it's given a non-event so we can skip this check here.

We still need to check the number of keys though, as:
* 0 keys would mean that the length of the array (and therefore the size of data->events) does not match the actual number of events

* more than one key does not make sense (at least for now, maybe one day this could be used to attach some sort of metadata to the event)

I need to fix the code to check we are not in the 0 keys situation, currently it would trigger a UB with `data->events[i]` not being initialized.

src/rt-app_parse_config.c
douglas-raillard-arm: My point is that this code cannot be removed, if someone is using old-style, this foreach needs to be somewhere. Silently ignoring part of the configuration file in order to save one iteration over a map that has a few tens of keys at most does not sound like a good deal. Even more considered the fact that in well-formed JSON files, the number of keys is bounded by the number of settings, so the only ones "paying a performance cost" are people writing a conf file that is ambiguous and will result in an error. (bearing in mind, we are parsing a configuration file, ahead of time of execution, in C, with an order of magnitude of a few tens or hundreds of keys in extreme cases).

On top of that, this can allow in the future to reclaim some event names for some settings. This sounds a bit crazy given that they have "verb" names for the most part, but the current syntax is lenient: we only match the prefix. This means that something like "synchronous-io" key is currently impossible to add, but would be possible with the "events" array as long as we forbid events outside of it when in use.

PR #111, opened 2021-02-05
Alpine linux fixes

PierreARM:
Some build errors prevented building rt-app in Alpine Linux. These modifications should allow building.

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -563,7 +563,7 @@ static int run_event(event_data_t *event, int dry_run,
 	case rtapp_yield:
 		{
 			log_debug("yield %d", event->count);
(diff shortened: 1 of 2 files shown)


credp:
Hi Vincent, we think this was uncontroversial, would you mind if I merge it?

vingu-linaro:
That's ok. I think that the only open question was about the difference in behavior. But looking at https://man7.org/linux/man-pages/man3/pthread_yield.3.html : 

CONFORMING TO         top
       This call is nonstandard, but present on several other systems.
       Use the standardized sched_yield(2) instead.

credp:
Excellent Vincent, Thanks

Issue #112, opened 2021-02-08
impossible to reset uclamp


douglas-raillard-arm:
rt-app currently reserves the -1 value as a special "don't modify" value. Unfortunately, -1 is now also a valid value for the clamps for the kernel, with the meaning "remove the clamp".

rt-app should free the -1 internal value (by using -2 or anything else) to let the user pass it to the kernel.

PR #113, opened 2021-02-08
uclamp: Allow to pass -1 to sched_setattr()

douglas-raillard-arm:
NOT TESTED YET

Since special value (u32)-1 removes the clamp for the task in the
kernel, free -1 special value from rt-app internals so that the user can
make use of it. Replace it with -2 instead.

fixes https://github.com/scheduler-tools/rt-app/issues/112


douglas-raillard-arm:
Now tested

PR #116, opened 2021-02-18
ftrace: Add phase_counter rtapp_loop field

douglas-raillard-arm:
Add a phase_counter field containing a strictly increasing per-thread
counter that is incremented at each phase loop. This allows to easily
match a given phase to the corresponding high level object in rt-app
configuration file builder such as lisa.wlgen.rta Python module.

With the strictly incrementing counter, the fact that the phases have
been folded into a minimal sequence with thread loop > 0 becomes an
implementation detail and will not impact later analysis of the trace,
even when it has been cropped to a small window that prevents counting
the number of phases that were present in the JSON file.

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -1136,7 +1136,7 @@ void *thread_body(void *arg)
 	timing_point_t tmp_timing;
 	unsigned int timings_size, timing_loop;
 	struct sched_attr attr;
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
Abandoning as the feature is unlikely to make it and workarounds exist

Issue #119, opened 2021-03-10
error with ftrace examples


amamory:
When running any example with ftrace enable, it causes the following error:

# rt-app examples/tutorial/example1.json
<error> Cannot write mark_fd: Bad file descriptor

This error happen when _tracing_on_ is not enabled before writing into _trace_marker_.

Looking at the following [line of code](https://github.com/scheduler-tools/rt-app/blob/665eececccea20b6cb55bfed3a2c02a7d3773a2a/src/rt-app.c#L1545):

> 		log_notice("configuring ftrace");
> 		strcpy(tmp, ft_data.debugfs);
> 		strcat(tmp, "/tracing/tracing_on");
> 		strcpy(tmp, ft_data.debugfs);
> 		strcat(tmp, "/tracing/trace_marker");
> 		ft_data.marker_fd = open(tmp, O_WRONLY);

As can be seen in this code, nothing is written into tracing_on, before writing into trace_marker.
There should be something equivalent to:

> echo 1 > /sys/kernel/debug/tracing/tracing_on
> echo 'something' > /sys/kernel/debug/tracing/trace_marker
> echo 0 > /sys/kernel/debug/tracing/tracing_on

Also, it seems that there is no place disabling the tracing.

amamory:
Pull #121 fix this as it first checks whether tracing is enabled outside the app scope.
For example, if the json file enables ftrace, it is expected that the user runs ftrace (i.e., set _tracing_on_ to '1' or _trace-cmd record_ ...) before running rt-app. If the user forget it, the rt-app will issue an error 'tracing is not enabled in file ....tracing/tracing_on'.

PR #120, opened 2021-03-10
partial fix for issue #119 related to ftrace

amamory:
a potential solution to issue #119, but it needs more testing by someone with more experience with ftrace.


amamory:
I have not realised that the program is assuming ftrace is executed outside the scope of the program. this pull can be ignored and i will post another one with an additional check to avoid the error msg related to #119.

PR #121, opened 2021-03-11
additional check when ftrace is enabled

amamory:
this check will avoid the error message reported in #119 .


vingu-linaro:
make sense
Thanks

PR #122, opened 2021-03-31
src: Fix incorrect use of cpu_set_t

douglas-raillard-arm:
cpu_set_t allocated dynamically with CPU_ALLOC need to be manipulated
with CPU_*_S() family of macro, otherwise it will be assumed that the
size is equal to sizeof(cpu_set_t), leading to undefined behaviours.
a


vingu-linaro:
As we have:
		data->cpusetsize = sizeof(cpu_set_t);
		data->cpuset = malloc(data->cpusetsize);
Isn't the assumption correct ?

douglas-raillard-arm:
That's not the only place where a cpuset is created:
https://github.com/scheduler-tools/rt-app/blob/master/src/rt-app.c#L844

Currently, rt-app crashes with a given input and when the JSON file name (as passed to CLI) is larger than 92 chars. The crash happens inside CPU_EQUAL_S, where it can be seen that the size that the libc is assuming differs from what is in the `cpusetsize` member.

In the details, the musl libc defines:
```
typedef struct cpu_set_t { unsigned long __bits[128/sizeof(long)]; } cpu_set_t;
```
So a cpu_set_t is 128 bytes long. When we use the dynamic allocation API, we end up with a much smaller object. The fixed-size API ends up using memcmp on with size=128 whereas the value are smaller, so it ends up reading garbage (and unmapped area when the stars are aligned).

douglas-raillard-arm:
Updated with a corrected call to CPU_SET_S (arguments were swapped) and a commit to use the fixed-size API instead of the variably-sized API.

credp (approved):
lgtm

credp:
@vingu-linaro I'm considering if it is appropriate for me to merge these changes - they look good to me, we need them because our clib triggers the issue, and we've been using them for a while now. Would you have any objection to me hitting the big green button?

Thanks!

vingu-linaro:
@credp Let me try to remember what was the open question for this PR

vingu-linaro:
patch 1 says: cpu_set_t allocated dynamically with CPU_ALLOC need to be manipulated
with CPU_*_S() family of macro
but patch 2 removes the only call to CPU_ALLOC and uses malloc(sizeof(cpu_set_t)) everywhere instead
also cpusetsize is now always set to sizeof(cpu_set_t)

So should we remove patch1 and only keeps patch2 ?

douglas-raillard-arm:
I think the reason I removed `CPU_ALLOC`  the data it's initialized with come from `cpu_set_t` anyway:
```
		cpu_set_t cpuset;
		unsigned int cpu_count;

		ret = pthread_getaffinity_np(pthread_self(),
						    sizeof(cpu_set_t), &cpuset);

...
		cpu_count = CPU_COUNT(&cpuset);
		data->def_cpu_data.cpusetsize = CPU_ALLOC_SIZE(cpu_count);
		data->def_cpu_data.cpuset = CPU_ALLOC(cpu_count);

```
So there is no point in using the dynamic size API, since by construction we will never get  a `CPU_COUNT()` that cannot be represented in `cpu_set_t`. 

With both patches, you end up with client code that does not care about what specific type of cpu set is being used. If you drop the first patch, you also drop the ability to easily switch back to a dynamically allocated cpu set (I suppose another API than pthread allows getting the cpu count no matter how big). And that would amount to shooting yourself in the foot:
https://git.musl-libc.org/cgit/musl/tree/include/sched.h#n127

Currently, rt-app will fail to handle more than 128 CPUs in proper static builds, so I wouldn't drop the code that will allow fixing that easily.

vingu-linaro:
> I think the reason I removed `CPU_ALLOC` the data it's initialized with come from `cpu_set_t` anyway:
> 
> ```
> 		cpu_set_t cpuset;
> 		unsigned int cpu_count;
> 
> 		ret = pthread_getaffinity_np(pthread_self(),
> 						    sizeof(cpu_set_t), &cpuset);
> 
> ...
> 		cpu_count = CPU_COUNT(&cpuset);
> 		data->def_cpu_data.cpusetsize = CPU_ALLOC_SIZE(cpu_count);
> 		data->def_cpu_data.cpuset = CPU_ALLOC(cpu_count);
> ```
> 
> So there is no point in using the dynamic size API in the current state of the code, since by construction we will never get a `CPU_COUNT()` that cannot be represented in `cpu_set_t`.
> 
> With both patches, you end up with client code that does not care about what specific type of cpu set is being used. If you drop the first patch, you also drop the ability to easily switch back to a dynamically allocated cpu set (I suppose another API than pthread allows getting the cpu count no matter how big). And that would amount to shooting yourself in the foot: https://git.musl-libc.org/cgit/musl/tree/include/sched.h#n127
> 
> Currently, rt-app will fail to handle more than 128 CPUs in proper static builds, so I wouldn't drop the code that will allow fixing that easily.

But all this is quite misleading: 
-Either you fix the system to use dynamically allocated cpu set
-Or you fix it to only use statically allocated.

src/rt-app.c
```diff
@@ -861,8 +856,12 @@ static void set_thread_affinity(thread_data_t *data, cpuset_data_t *cpu_data)
 	if (actual_cpu_data->cpuset == NULL)
 		actual_cpu_data = &data->def_cpu_data;
 
-	if (!CPU_EQUAL(actual_cpu_data->cpuset, data->curr_cpu_data->cpuset))
-	{
+	if (
+	    actual_cpu_data->cpusetsize != data->curr_cpu_data->cpusetsize ||
```
vingu-linaro, line 860: Witht this PR, cpusetsize is always set to sizeof(cpu_set_t) so this test is useless

douglas-raillard-arm:
> But all this is quite misleading:
-Either you fix the system to use dynamically allocated cpu set
-Or you fix it to only use statically allocated.

I fix the current segfault, make the code agree with itself (I did not introduce `cpusetsize` it was already there but used only half of the time) while preparing the ground to do the next step. Feel free to drop the PR if fixing the segfault is not enough.

vingu-linaro:
> > But all this is quite misleading:
> > -Either you fix the system to use dynamically allocated cpu set
> > -Or you fix it to only use statically allocated.
> 
> I fix the current segfault, make the code agree with itself (I did not introduce `cpusetsize` it was already there but used only half of the time) while preparing the ground to do the next step. Feel free to drop the PR if fixing the segfault is not enough.

I'm not saying the PR is irrelevant, I'm saying that patch 1 is irrelevant because patch 2 removes the code that patch1 wants to fixes so patch 1 is useless. That's why I asked you to keep patch 2 and remove patch 1

douglas-raillard-arm:
Tested the branch in its current state including the last 2 patches, they build and seem to work fine:
* Tasks are pinned to the correct CPUs
* calibration works well (one value for each CPU, correctly detected)

src/rt-app_parse_config.c
```diff
@@ -763,17 +763,17 @@ static void parse_cpuset_data(struct json_object *obj, cpuset_data_t *data)
...
 		data->cpuset = malloc(data->cpusetsize);
-		CPU_ZERO(data->cpuset);
+		CPU_ZERO_S(data->cpusetsize, data->cpuset);
 		for (i = 0; i < json_object_array_length(cpuset_obj); i++) {
 			cpu = json_object_array_get_idx(cpuset_obj, i);
 			cpu_idx = json_object_get_int(cpu);
 			if (cpu_idx > max_cpu) {
 				log_critical(PIN2 "Invalid cpu %u in cpuset %s", cpu_idx, data->cpuset_str);
-				free(data->cpuset);
+				CPU_FREE(data->cpuset);
```
vingu-linaro, line 772: we have data->cpuset = malloc(data->cpusetsize); few lines above so why replacing free() by CPU_FREE() ?

we should use the same alloc function everywhere: CPU_ALLOC(MAX_CPUS); and CPU_ALLOC_SIZE(MAX_CPUS);

vingu-linaro:
Please use everywhere 
cpusetsize = CPU_ALLOC_SIZE(MAX_CPUS);
cpuset = CPU_ALLOC(MAX_CPUS);CPU_ALLOC(MAX_CPUS); and (MAX_CPUS);

and squash the first 3 commits in a single one

PR #126, opened 2022-07-06
ftrace_log: Add override for tracefs location

credp:
It's possible to mount tracefs in other locations - the Google Pixel 6,
for example, mounts tracefs at /sys/kernel/tracing rather than the
typical location inside debugfs at /sys/kernel/debug/tracing and disallows
debugfs from being mounted when running any GKI kernel.

This change allows configurations to select the location where rt-app
will attempt to configure tracing when enabled. If it is not supplied,
then the intended behaviour is that the existing default should be used.

There should not be any impact on existing json files running on existing
systems.

Signed-off-by: "Chris Redpath" <chris.redpath@arm.com>

### DIFF
--- a/doc/tutorial.txt
+++ b/doc/tutorial.txt
@@ -92,6 +92,11 @@ The category names currently supported are:
 
 Default value is "none".
 
+* ftracedir : String. Prefix used to find tracefs. Defaults to
+"/sys/kernel/debug" if not specified. In that case, we expect to find tracing
+controls inside "/sys/kernel/debug/tracing/". Has no effect if ftrace is not
+enabled.
+
 * gnuplot : Boolean. If True, it will create a gnu plot compatible file for
 each threads (see gnuplot section for more details). Default value is False.
 
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -64,7 +64,6 @@ static pthread_mutex_t joining_mutex;
 static pthread_mutex_t fork_mutex;
 
 static ftrace_data_t ft_data = {
-	.debugfs = "/sys/kernel/debug",
 	.marker_fd = -1,
 };
 
@@ -1542,7 +1541,7 @@ int main(int argc, char* argv[])
 	if (ftrace_level != FTRACE_NONE) {
 		log_notice("configuring ftrace");
 		// check if tracing is enabled
-		strcpy(tmp, ft_data.debugfs);
+		strcpy(tmp, opts.ftracedir);
 		strcat(tmp, "/tracing/tracing_on");
 		int ftrace_f = open(tmp, O_RDONLY);
 		if (ftrace_f < 0){
@@ -1557,7 +1556,7 @@ int main(int argc, char* argv[])
 		}
 		close(ftrace_f);
 		// set the marker
-		strcpy(tmp, ft_data.debugfs);
+		strcpy(tmp, opts.ftracedir);
 		strcat(tmp, "/tracing/trace_marker");
 		ft_data.marker_fd = open(tmp, O_WRONLY);
 		if (ft_data.marker_fd < 0) {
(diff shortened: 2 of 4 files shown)


douglas-raillard-arm:
@credp it builds but there is a warning to fix:
```
rt-app_parse_config.c:1210:33: warning: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
 1210 |                 opts->ftracedir = ftrace_dir();
      |                                 ^
```

credp:
> @credp it builds but there is a warning to fix:
> 
> ```
> rt-app_parse_config.c:1210:33: warning: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
>  1210 |                 opts->ftracedir = ftrace_dir();
>       |                                 ^
> ```

Ooh, thanks. I changed the ftrace_dir to use strdup instead of using a static string buffer and forgot to fix this bit up. Will repost

doc/tutorial.txt
```diff
@@ -92,6 +92,11 @@ The category names currently supported are:
 
 Default value is "none".
 
+* ftracedir : String. Prefix used to find tracefs. Defaults to
```
douglas-raillard-arm, line 95: Once the "tracing subfolder issue" is fixed, it should probably be simplified into:
ftracedir: String: Mount point of tracefs, or equivalent in older systems. Defaults to "/sys/kernel/debug/tracing"

credp: Yep, I tidied that up a bit.

src/rt-app.c
```diff
@@ -64,7 +64,6 @@ static pthread_mutex_t joining_mutex;
 static pthread_mutex_t fork_mutex;
 
 static ftrace_data_t ft_data = {
-	.debugfs = "/sys/kernel/debug",
```
vingu-linaro, line 67: Could you update _ftrace_data_t and remove the now unused field debugfs ?
typedef struct _ftrace_data_t {
	char *debugfs;
	int trace_fd;
	int marker_fd;
} ftrace_data_t;

Also the trace_fd doesn't seem to be used anywhere which let the struct with only marker_fd. could be good to move it in opt struct as well

src/rt-app_utils.c
```diff
@@ -267,6 +268,60 @@ resource_to_string(resource_t resource, char *resource_name)
...
+		tmp = malloc(len);
+		if(!tmp)
+			return NULL;
+		strcpy(tmp, path);
+		free(path);
+		strcat(tmp, "/tracing");
+		return tmp;
+	}
+
+	return strdup(DEFAULT_FTRACE_DIR);
```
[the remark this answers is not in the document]
credp: If I return NULL here, it means that if the mtab code fails we cannot use ftrace unless we provide a path in the global section. That's probably OK, I don't mind that if you dont.

src/rt-app.c
credp: @vingu-linaro thanks :) I will rebase with trace_fd removed in the first patch and I'll move marker_fd out. Incidentally, would you mind if I squash the patches together?

vingu-linaro: I'm fine if you squash patches together

src/rt-app_utils.c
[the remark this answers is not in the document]
vingu-linaro: I wondering if there is a usecase where mtab would fail but the default path could work

src/rt-app.c
credp: Cool

src/rt-app_utils.c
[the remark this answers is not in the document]
credp: I don't think so, and if that does happen you can always add ftracedir to your global options file to force it to a specific location. I'll push a new squashed rev up once I've tested it a bit.

PR #127, opened 2023-01-30
Remove mentioning of with-deadline in documentation

cloehle:
Support for SCHED_DEADLINE is the default since
commit af279433a58e ("Always build with deadline").
The parameter does not do anything anymore, so remove it.

Signed-off-by: Christian Loehle <cloehle@posteo.de>


vingu-linaro:
Thanks for the cleanup

PR #129, opened 2023-09-14
Update build documentation

cloehle:
numactl has switched to cmake and therefore the build instructions had to be updated. Other minor adaptions were done, too.

numactl when used by file path needs to be in the numactl directory, so that was updated. Furthermore the git clone was removed to be in line with the other github link.


vingu-linaro:
> numactl has switched to cmake and therefore the build instructions had to be updated. Other minor adaptions were done, too.
> 
> numactl when used by file path needs to be in the numactl directory, so that was updated. Furthermore the git clone was removed to be in line with the other github link.

After checking json-c and numactl repo, it seems that json-c moved to cmake and not numactl

PR #131, opened 2024-02-20
Bjdooks/fixes 20feb2024

bjdooks-ct:
Some bugfixes on error return from calibration, and the wrong error code from sched_setattr.
Fix a indentation issue on rtapp_sleep


bjdooks-ct:
pushed wrong branch, has a typo still in the code

PR #133, opened 2024-07-07
Add cgroup v2 support

uran0sH:


uran0sH:
@vingu-linaro if you have time, plz review this pr. thanks

src/rt-app_taskgroups.c
```diff
@@ -373,7 +402,14 @@ static int cgroup_attach_task(char *name)
 	FILE *tasks;
 	size_t size;
 
-	file = strcmp(name, "/") ? "/tasks" : "tasks";
+	if (ctrl.version == 1) {
+		file = strcmp(name, "/") ? "/tasks" : "tasks";
+	}
+
+	if (ctrl.version == 2) {
+		file = strcmp(name, "/") ? "/cgroup.procs" : "cgroup.procs";
```
Werkov, line 410: Notice this moves whole process (threadgroup in contrast to `tasks` that works with threads).
If this should support also thread granularity, `cgroup.type` should be used too (see e.g. cgroups(7)).

uran0sH: I have changed it to use "cgroup.threads". Plz review again.

src/rt-app_taskgroups.c
```diff
@@ -157,23 +161,34 @@ static int cgroup_get_cpu_controller_mount_point(void)
...
+			}
 
-		if (!hasmntopt(ent, "cpu"))
-			continue;
+			strcpy(ctrl.mount_point, ent->mnt_dir);
+			log_debug(PIN "cgroup cpu controller mountpoint [%s] found", ent->mnt_dir);
+			ret = 0;
+			break;
+		} else if (!strcmp(ent->mnt_type, "cgroup2")) {
+			ctrl.mount_point = malloc(strlen(ent->mnt_dir) + 1);
```
vingu-linaro, line 180: I have seen system with both cgroup (for cpu) and cgroup2. I'm not sure that getmntent() returns a sorted listed of mount point which means that if you check for cgroup2 before cgroup you will return whereas the cpu controller is in cgroup. you need to check in cgroup.controllers that cpu controller is available

src/rt-app_taskgroups.c
```diff
@@ -253,6 +268,60 @@ static int cgroup_mkdir(const char *name, int *offset)
...
+
+			path_size = strlen(path) + strlen(PROCS) + 2;
+			cgroup_path = malloc(path_size);
+			snprintf(cgroup_path, path_size, "%s/%s", path, PROCS);
+			cgroup_file = fopen(cgroup_path, "we");
+			if (!cgroup_file) {
+				perror("fopen");
+				goto error;
+			}
+			if (fprintf(cgroup_file, "%d", getpid()) < 0) {
```
vingu-linaro, line 315: Why are you putting the full rt-app process in the newly created group ? The end result is that all threads will be in the last created cgroup by default even those which don't set a "taskgroup" field

src/rt-app_taskgroups.c
```diff
@@ -157,23 +161,34 @@ static int cgroup_get_cpu_controller_mount_point(void)
...
 
 	while (ent = getmntent(mounts)) {
-		if (strcmp(ent->mnt_type, "cgroup"))
-			continue;
+		if (!strcmp(ent->mnt_type, "cgroup")) {
+			if (!hasmntopt(ent, "cpu"))
+				continue;
+
+			ctrl.mount_point = malloc(strlen(ent->mnt_dir) + 1);
+			ctrl.version = 1;
```
vingu-linaro, line 169: you should set ctrl.version after checking that the malloc succeed

src/rt-app_taskgroups.c
Werkov: Additionally, wouldn't this drag the process through all created groups (top-down)?

src/rt-app_taskgroups.c
```diff
@@ -253,6 +268,60 @@ static int cgroup_mkdir(const char *name, int *offset)
...
+			
+			path_size = strlen(path) + strlen(TYPE) + 2;
+			cgroup_path = malloc(path_size);
+			snprintf(cgroup_path, path_size, "%s/%s", path, TYPE);
+			cgroup_file = fopen(cgroup_path, "we");
+			if (!cgroup_file) {
+				perror("fopen");
+				goto error;
+			}	
+			if (fprintf(cgroup_file, "threaded") < 0) {
```
Werkov, line 297: Does this work for you?

It may but it may also break if there's "domain" controller (e.g. memory) on the ancestry path. Perhaps you can live with that, alternatively threaded mode could only be enabled on the target cgroup (provided the leaf doesn't have "domain" controller too).

uran0sH:
In summary, I should only enable the threaded mode on the target cgroup type(like Werkov said that I also need to check the domain in other controllers) and write the tid to the cgroup.procs? @vingu-linaro @Werkov

src/rt-app_taskgroups.c
```diff
@@ -373,7 +435,35 @@ static int cgroup_attach_task(char *name)
...
+				perror("fprintf");
+				goto error;
+			}
+			if (fclose(cgroup_file)) {
+				perror("fclose");
+				goto error;
+			}
+			free(cgroup_path);
+		}
+		file = strcmp(name, "/") ? "/cgroup.procs" : "cgroup.procs";
```
Werkov, line 464: Even when you switch the cgroup to `threaded` mode a write through `cgroup.procs` would migrate whole process (threadgroup) into target cgroup. The difference may not be visible with single-threaded processes. I assume rt-app is supposed to work with invididual threads of  a process.

OTOH, you may not always write only into `cgroup.threads`, e.g. because you can't pick a single thread of a (foreign) process and move it between cgroups that are not members of the same threaded subtree. (Not sure if this is the mode how rt-app operates.)

Werkov (commented):
Your code seem to work in some situations (single threaded processes) but could be broken in some other (noted in a comment). There may be no tests in rt-app that would capture this so it may only become visible when used by more users :)

src/rt-app_taskgroups.c
vingu-linaro: rt-app is a single process that creates several threads so you should use cgroup.threads instead of cgroup.procs to move each thread in different cgroup as mentioned by @Werkov.

src/rt-app_taskgroups.c
```diff
@@ -373,7 +435,35 @@ static int cgroup_attach_task(char *name)
 	FILE *tasks;
 	size_t size;
 
-	file = strcmp(name, "/") ? "/tasks" : "tasks";
+	if (ctrl.version == 1) {
+		file = strcmp(name, "/") ? "/tasks" : "tasks";
```
Werkov, line 439: I wonder it this special casing for root cgrp is worth it (here and other places).

```diff
-sprintf(path, "%s%s%s", ctrl.mount_point, name, file);
+sprintf(path, "%s%s/%s", ctrl.mount_point, name, file);
```
with possibly double `//` being resolved by open.

src/rt-app_taskgroups.c
```diff
@@ -393,9 +483,11 @@ static int cgroup_attach_task(char *name)
 		goto error;
 	}
 
-	if (fclose(tasks)) {
-		perror("fclose");
-		goto error;
+	if (ctrl.version == 1) {
```
Werkov, line 486: v2 file deserves to be closed here too, no?

src/rt-app_taskgroups.c
```diff
@@ -373,7 +435,35 @@ static int cgroup_attach_task(char *name)
...
+			FILE* cgroup_file;
+
+			snprintf(cgroup_path, path_size, "%s/%s/%s", ctrl.mount_point, name, TYPE);
+			cgroup_file = fopen(cgroup_path, "we");
+			if (!cgroup_file) {
+				perror("open");
+				goto error;
+			}	
+			if (fprintf(cgroup_file, "threaded") < 0) {
+				perror("fprintf");
```
Werkov, line 455: cgroup_file not closed here

Werkov: Also some better message may ease future debugging, like `cgroup.type <- threaded failed`.

Werkov (commented):
This could work.
Added some stylistic/resource tracking remarks.

PR #136, opened 2025-03-08
Fix build with GNU libc 2.41

aurel32:
Starting with GNU libc 2.41, the sched_getattr and sched_setattr functions, and the sched_attr structure are provided in sched.h. This causes a conflict with the definition in libdl.

The configure script already checks for the sched_setattr function, but then the logic to compile and link libdl was wrong, as it is not possible to directly check the HAVE_SCHED_SETATTR with AM_CONDITIONAL. In addition the dl_syscalls.h include file needs to be guarded, and a few include files that were indirectly included by this file not to be done explictly.


vingu-linaro:
I'm still on 2.40 and I tested your change with it.
Your change looks good for me

PR #138, opened 2025-06-22
Cleanups and fixes

jan-kiszka:
Some smaller fixes and some cleanups. Some features will follow on top later, see https://github.com/scheduler-tools/rt-app/issues/137.

PR #141, opened 2025-08-29
Lockpages, events and timer assorted fixes

jlelli:
Move lockpages before setscheduler. Add additional information for lockpages and setscheduler. Reset t_first right before changing scheduling policy so that timers (and in-kernel accounting) is not skewed.

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -1141,6 +1141,10 @@ static void set_thread_param(thread_data_t *data, sched_data_t *sched_data)
 	}
 
 	data->curr_sched_data = sched_data;
+
+	log_ftrace(ft_data.marker_fd, FTRACE_TASK,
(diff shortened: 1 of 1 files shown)


vingu-linaro:
The mlock reordering modifies (https://github.com/scheduler-tools/rt-app/pull/138/commits/5890dd37cacc7693a2790d833ce38ebf78835055). Locking pages whatever the scheduling policy makes sense to me but I'd like to know what @jan-kiszka thinks about it if it was only because of the [data->lock_pages = 0](https://github.com/scheduler-tools/rt-app/pull/141/commits/97a61ff23aa9386cff7ee3f93623ef57d47dddf4#diff-1043a07ad3a161748bb92ec659dab456fe6d4f031af78ab29cea518ec63fe0cbL1133)

PR #147, opened 2026-04-13
rt-app: Add memrun event with read, write, and chase workload types

sarav:
Add a new "memrun" event that supports three memory workload types beyond the existing "mem" (memset) and "iorun" (write syscall):

  - "write": byte-at-a-time volatile stores over a sized buffer
  - "read": byte-at-a-time volatile reads over a sized buffer
  - "chase": pointer-chasing over a linked list for memory latency measurement, inspired by lmbench's lat_mem_rd

The read and write implementations use nested loops to avoid modulo overhead — the outer loop counts full buffer passes, the inner loop does sequential access. This gives symmetric read/write performance (unlike the legacy "mem" event which uses vectorized memset).

The chase workload supports configurable buffer size, stride, and pattern. The "random" pattern uses bit-reversed pointer ordering (from lmbench) to defeat hardware prefetchers, while "sequential" uses a simple stride-based chain. Buffer size controls which cache level is stressed.

JSON syntax:
  "memrun": { "type": "chase", "size": 262144, "stride": 64,
              "pattern": "random", "count": 50000000 }
  "memrun": { "type": "read", "size": 10485760, "count": 10485760 }
  "memrun": { "type": "write", "size": 10485760, "count": 10485760 }

The existing "mem" integer event is unchanged for backward compatibility.

Verified on ARM64 with simpleperf showing clear differentiation across workload types in instruction and bus-access stats.

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -704,6 +704,18 @@ static int run_event(event_data_t *event, int dry_run,
 			pthread_mutex_unlock(&fork_mutex);
 		}
 		break;
+	case rtapp_sem_post:
+		{
+			log_debug("sem_post %s", rdata->name);
+			sem_post(&rdata->res.sem.obj);
+		}
+		break;
+	case rtapp_sem_wait:
+		{
+			log_debug("sem_wait %s", rdata->name);
+			sem_wait(&rdata->res.sem.obj);
+		}
+		break;
 	default:
 		break;
 	}
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -409,6 +409,58 @@ static void memload(unsigned long count, struct _rtapp_iomem_buf *iomem)
 	}
 }
 
+static void memread(unsigned long count, struct _rtapp_iomem_buf *iomem)
+{
+	volatile char *p = (volatile char *)iomem->ptr;
+	unsigned long buf_size = iomem->size;
+	unsigned long passes = count / buf_size;
+	unsigned long remainder = count % buf_size;
+	unsigned long i, j;
+
+	/*
+	 * p is volatile, so each access is a side effect that the compiler
+	 * must emit even when the value is discarded. No accumulator or ALU
+	 * op needed to keep the loads alive.
+	 */
+	for (i = 0; i < passes; i++)
+		for (j = 0; j < buf_size; j++)
+			(void)p[j];
+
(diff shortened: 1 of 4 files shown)


src/rt-app.c
```diff
@@ -1615,6 +1748,13 @@ int main(int argc, char* argv[])
 	initialize_cgroups();
 	add_cgroups();
 
+	/* Initialize mem_chase pointer chains for all resources */
+	for (i = 0; i < opts.resources->nresources; i++) {
```
vingu-linaro: Why this is not part of init_resources ?

vingu-linaro: we want to keep all resources init in same place

sarav: Yes this should be moved. But should I move this to parse_task_event_data or does it have to be init_resources_data()? Problem with the latter is that we don't have the event details (buf size, stride, pattern, etc) in there.

So I either have to put it in parse... or pass these args (union of per resource args) into init_resource_data().

Which approach do you prefer?

vingu-linaro: Passing the union of per resource arg into init_resource_data is the cleanest solution IMO

vingu-linaro:
Also I forgot to mention that you should update the doc/tutorial.txt with the description of the new events

src/rt-app.c
sarav: Ok let me do that

sarav:
Found a bunch of bugs in the previous implementation when I tested esoteric combinations of these memrun events. Fixed them all.

* Memory leaks in memrun read/write.
* Wrong buffers when the names collide but the params are different.
* Unnecessary math in memrun read slowing it down. 
* Refactored to addresses comments.

vingu-linaro:
The code that adds semaphore events, looks good to me but you still have to update tutorial.txt with the description of these new events
I'm still reviewing the patch that adds the new mem events and in particular the mem chase event

sarav:
Updated the docs. Added a 3rd commit to allow memrun to share buffers across threads with documentation for it and use case examples for it.

vingu-linaro:
@sarav, the patchset looks good to me. Let see if others have comment otherwise I will merge it beg of the week

Thanks

sarav:
Thanks @vingu-linaro

PR #149, opened 2026-04-30
rt-app: Allow to set only nice or runtime value for SCHED_OTHER/BATCH…

deggeman:
… tasks

sa_params.sched_flags = SCHED_FLAG_KEEP_PARAMS can't be used to keep the current se.slice value for a task in case it is not specified in its rt-app profile since in this case the nice value which might be in the profile gets ignored too:

Kernel-side sched_setattr retrieves the parameters from the task in case SCHED_FLAG_KEEP_PARAMS is set:

  SYSCALL_DEFINE3(sched_setattr, ...) does:

  if (attr.sched_flags & SCHED_FLAG_KEEP_PARAMS)
    get_params(p, &attr, 0);

So in case either sched_data's prio or runtime is not set, retrieve the task parameters from the kernel to be able to set the value in case it is not specified in sched_data.

Make sure that the sched_data->prio == THREAD_PRIORITY_UNCHANGED can reach __set_thread_sched_other_attrs().

### DIFF
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -948,12 +948,17 @@ static bool __set_thread_policy_priority(thread_data_t *data,
 static void __set_thread_sched_other_attrs(thread_data_t *data,
 					   sched_data_t *sched_data)
 {
-	int ret;
-	struct sched_attr sa_params = {0};
+	int ret, prio_unchanged;
+	struct sched_attr sa_params = {0}, _sa_params = {0};
 	unsigned int flags = 0;
 	pid_t tid;
 
-	if (sched_data->prio > 19 || sched_data->prio < -20) {
+	prio_unchanged = sched_data->prio == THREAD_PRIORITY_UNCHANGED;
+
+	if (prio_unchanged && !sched_data->runtime)
+		return;
+
+	if (!prio_unchanged && (sched_data->prio > 19 || sched_data->prio < -20)) {
 		log_critical("[%d] sched_setattr %d nice invalid. "
 			     "Valid between -20 and 19",
 			     data->ind, sched_data->prio);
@@ -968,22 +973,35 @@ static void __set_thread_sched_other_attrs(thread_data_t *data,
 		   policy_to_string(sched_data->policy), sched_data->prio,
 		   sched_data->runtime);
 
+	if (prio_unchanged || !sched_data->runtime) {
+		_sa_params.size = sizeof(_sa_params);
+
+		if (sched_getattr(0, &_sa_params, sizeof(_sa_params), 0) == -1) {
+			perror("sched_getattr: failed to get SCHED_OTHER attributes");
+			exit(EXIT_FAILURE);
+		}
+	}
+
 	tid = gettid();
 	sa_params.size = sizeof(struct sched_attr);
 	sa_params.sched_policy = sched_data->policy;
 	sa_params.sched_priority = __sched_priority(data, sched_data);
+
 	/* In the CFS case, sched_data->prio is the NICE value. */
-	sa_params.sched_nice = sched_data->prio;
+	if (!prio_unchanged)
+		sa_params.sched_nice = sched_data->prio;
+	else
+		sa_params.sched_nice = _sa_params.sched_nice;
+
 	/*
 	 * Since Linux v6.12 it is possible to request a custom slice length
 	 * for tasks using a fair.c policy (other than SCHED_IDLE) via
 	 * sched_attr::sched_runtime.
 	 */
-
-	if(sched_data->runtime)
+	if (sched_data->runtime)
 		sa_params.sched_runtime = sched_data->runtime;
 	else
-		sa_params.sched_flags = SCHED_FLAG_KEEP_PARAMS;
+		sa_params.sched_runtime = _sa_params.sched_runtime;
 
 	ret = sched_setattr(tid, &sa_params, flags);
 	if (ret) {
@@ -997,9 +1015,6 @@ static void __set_thread_sched_other_attrs(thread_data_t *data,
 
 static void _set_thread_cfs(thread_data_t *data, sched_data_t *sched_data)
 {
-	/* Priority unchanged => Policy unchanged */
-	if (sched_data->prio == THREAD_PRIORITY_UNCHANGED)
-		return;
 	/*
 	 * In the CFS case, sched_data->prio is the NICE value. As long as the
 	 * policy hasn't changed, there's no need to call
--- a/src/rt-app.c
+++ b/src/rt-app.c
@@ -111,7 +111,7 @@ static void thread_data_set_unique_name(thread_data_t *tdata, int nforks)
 		tdata->name = strdup(tdata->name);
 	}
 
-	log_notice("thread_data_set_unique_name %d %s", tdata->ind, tdata->name);
+	log_notice("[%d] set unique thread name %s", tdata->ind, tdata->name);
 }
 
 /*
@@ -911,20 +911,6 @@ static int __sched_priority(thread_data_t *data, sched_data_t *sched_data)
 	 return 0;
 }
 
-
-static void __log_policy_priority_change(thread_data_t *data,
-					 sched_data_t *sched_data)
-{
-	log_debug("[%d] setting scheduler %s priority %d", data->ind,
-		  policy_to_string(sched_data->policy),
-		  sched_data->prio);
-
-	log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
-		   "rtapp_attrs: event=policy policy=%s prio=%d",
-		   policy_to_string(sched_data->policy),
-		   sched_data->prio);
-}
-
 static bool __set_thread_policy_priority(thread_data_t *data,
 					 sched_data_t *sched_data)
 {
@@ -936,6 +922,7 @@ static bool __set_thread_policy_priority(thread_data_t *data,
 	ret = pthread_setschedparam(pthread_self(),
 				    sched_data->policy,
 				    &param);
+
 	if (ret) {
 		log_critical("[%d] pthread_setschedparam returned %d",
 			     data->ind, ret);
@@ -943,6 +930,13 @@ static bool __set_thread_policy_priority(thread_data_t *data,
 		perror("pthread_setschedparam");
 		exit(EXIT_FAILURE);
 	}
+
+	log_debug("[%d] setting scheduler %s priority %d", data->ind,
+		  policy_to_string(sched_data->policy), param.sched_priority);
+
+	log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+		   "rtapp_attrs: event=policy policy=%s prio=%d",
+		   policy_to_string(sched_data->policy), param.sched_priority);
 }
 
 static void __set_thread_sched_other_attrs(thread_data_t *data,
@@ -964,14 +958,6 @@ static void __set_thread_sched_other_attrs(thread_data_t *data,
 			     data->ind, sched_data->prio);
 		exit(EXIT_FAILURE);
 	}
-
-	log_debug("[%d] setting scheduler %s nice=%d runtime=%lu",
-		  data->ind, policy_to_string(sched_data->policy),
-		  sched_data->prio, sched_data->runtime);
-	log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
-		   "rtapp_attrs: event=policy policy=%s nice=%d runtime=%lu",
-		   policy_to_string(sched_data->policy), sched_data->prio,
-		   sched_data->runtime);
 
 	if (prio_unchanged || !sched_data->runtime) {
 		_sa_params.size = sizeof(_sa_params);
@@ -1011,6 +997,15 @@ static void __set_thread_sched_other_attrs(thread_data_t *data,
 		perror("sched_setattr: failed to set SCHED_OTHER attributes");
 		exit(EXIT_FAILURE);
 	}
+
+	log_debug("[%d] setting scheduler %s nice=%d runtime=%llu",
+		  data->ind, policy_to_string(sched_data->policy),
+		  sa_params.sched_nice, sa_params.sched_runtime);
+
+	log_ftrace(ft_data.marker_fd, FTRACE_ATTRS,
+		   "rtapp_attrs: event=policy policy=%s nice=%d runtime=%llu",
+		   policy_to_string(sched_data->policy), sa_params.sched_nice,
+		   sa_params.sched_runtime);
 }
 
 static void _set_thread_cfs(thread_data_t *data, sched_data_t *sched_data)
@@ -1029,8 +1024,6 @@ static void _set_thread_cfs(thread_data_t *data, sched_data_t *sched_data)
 
 	if (sched_data->policy == other || sched_data->policy == batch)
 		__set_thread_sched_other_attrs(data, sched_data);
-
-	__log_policy_priority_change(data, sched_data);
 }
 
 static void _set_thread_rt(thread_data_t *data, sched_data_t *sched_data)
@@ -1040,7 +1033,6 @@ static void _set_thread_rt(thread_data_t *data, sched_data_t *sched_data)
 		return;
 
 	__set_thread_policy_priority(data, sched_data);
-	__log_policy_priority_change(data, sched_data);
 }
 
 /* deadline can't rely on the default __set_thread_policy_priority */
@@ -1221,7 +1213,7 @@ void *thread_body(void *arg)
 
 	t_first = t_zero;
 
-	log_notice("[%d] starting thread ...\n", data->ind);
+	log_notice("[%d] starting thread %s", data->ind, data->name);
 
 	if (opts.logsize)
 		fprintf(data->log_handler, "%s %8s %8s %8s %15s %15s %15s %10s %10s %10s %10s\n",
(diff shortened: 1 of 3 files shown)


deggeman:
Only lightly tested with those rt-app json files on my Arm64 VM:

`---

{
        "tasks": {
                "t0": {
                        "loop" : -1,
                        "run" : 200,
                        "priority" : -10,
                },
        },
        "global": {
                "calibration" : 14,
        }
}

---

{
        "tasks": {
                "t0": {
                        "loop" : -1,
                        "run" : 200,
                        "dl-runtime" : 90000
                }
        },
        "global": {
                "calibration" : 14
        }
}

---

{
        "tasks": {
                "t0": {
                        "loop" : -1,
                        "phases": {
                                "p0" : {
                                        "run" : 100,
                                        "dl-runtime" : 40000,
                                },
                                "p1" : {
                                        "run" : 200,
                                        "dl-runtime" : 50000,
                                },
                        },
                }
        },
        "global": {
                "calibration" : 14
        }
}

---

{
        "tasks": {
                "t0": {
                        "loop" : -1,
                        "phases": {
                                "p0" : {
                                        "run" : 100,
                                        "priority" : -1,
                                },
                                "p1" : {
                                        "run" : 200,
                                        "priority" : -2,
                                },
                        },
                }
        },
        "global": {
                "calibration" : 14
        }
}

---

{
        "tasks": {
                "t0": {
                        "loop" : -1,
                        "phases": {
                                "p0" : {
                                        "run" : 100,
                                        "dl-runtime" : 40000,
                                },
                                "p1" : {
                                        "run" : 200,
                                        "dl-runtime" : 50000,
                                },
                        },
                }
        },
        "global": {
                "calibration" : 14,
                "default_policy": "SCHED_IDLE",
        }
}`

petretudor-arm:
> {
> "tasks": {
> "t0": {
> "loop" : -1,
> "phases": {
> "p0" : {
> "run" : 100,
> "dl-runtime" : 40000,
> },
> "p1" : {
> "run" : 200,
> "dl-runtime" : 50000,
> },
> },
> }
> },
> "global": {
> "calibration" : 14
> }
> }

With this test on a arm64 qemu setup I get these logs at the start of each phase:
```
[rt-app] <debug> [0] setting scheduler SCHED_OTHER nice=2147483647 runtime=40000000
[rt-app] <debug> [0] setting scheduler SCHED_OTHER priority 2147483647
```
In this case the nice value is clamped at the end of `sched_copy_attr()` [[1]] so we don't get the system value. This is what I was trying to say on #148.

To confirm this I added

```C
if (sched_getattr(0, &_sa_params, sizeof(_sa_params), 0) == -1)  {
        perror("sched_getattr: failed to get SCHED_OTHER attributes");
	        exit(EXIT_FAILURE);
}
log_debug("[%d] proceeding with nice=%d runtime=%llu", data->ind, _sa_params.sched_nice, _sa_params.sched_runtime);
```

right at the end of `__set_thread_sched_other_attrs()` and got

```
[rt-app] <debug> [0] setting scheduler SCHED_OTHER nice=2147483647 runtime=40000000
[rt-app] <debug> [0] proceeding with nice=19 runtime=40000000
[rt-app] <debug> [0] setting scheduler SCHED_OTHER priority 2147483647
```

[1]: https://elixir.bootlin.com/linux/v7.0/source/kernel/sched/syscalls.c#L872

petretudor-arm:
> {
> "tasks": {
> "t0": {
> "loop" : -1,
> "run" : 200,
> "dl-runtime" : 90000
> }
> },
> "global": {
> "calibration" : 14
> }
> }

With this test, THREAD_PRIORITY_UNCHANGED never reaches `__set_thread_sched_other_attrs()` because the lack of a phases object will create a single NULL phase `sched_data` in the thread's `thread_data_t` which gets ignored by `thread_set_param()`:
```C
		/* There is no "phases" object which means that thread and phase will
		 * use same scheduling parameters. But thread object looks for default
		 * value when parameters are not defined whereas phase doesn't.
		 * We remove phase's scheduling policy which is a subset of thread's one
		 */
		free(data->phases[0].sched_data);
		data->phases[0].sched_data = NULL;
```

A thread's default (initial) priority cannot be `same` because in `parse_task_data()` we have
```C
	data->sched_data = parse_sched_data(obj, opts->policy);
```
and the `opts->policy` comes from `parse_global` which sets it to SCHED_OTHER, so again we don't get the system default for prio.

If you add
```C
log_debug("[%d] set_thread_param() using policy=%d prio=%d runtime=%lu", data->ind, sched_data->policy, sched_data->prio, sched_data->runtime);
```
after the NULL check in `set_thread_param()` you can see the message only appears once in the logs, when setting the thread's initial scheduling parameters. And then the phases will not overwrite those with system values.

deggeman:
[rt-app] <debug> [0] setting scheduler SCHED_OTHER nice=2147483647 runtime=40000000
[rt-app] <debug> [0] setting scheduler SCHED_OTHER priority 2147483647

I think this is happening since we use INT_MAX forTHREAD_PRIORITY_UNCHANGED and not -1.

[rt-app] <debug> [0] proceeding with nice=19 runtime=40000000

This is a bug which can be fixed by:

@@ -984,7 +984,7 @@ static void __set_thread_sched_other_attrs(thread_data_t *data,
        sa_params.sched_priority = __sched_priority(data, sched_data);
 
        /* In the CFS case, sched_data->prio is the NICE value. */
-       if (sched_data->prio)
+       if (sched_data->prio != THREAD_PRIORITY_UNCHANGED)
                sa_params.sched_nice = sched_data->prio;
        else
                sa_params.sched_nice = _sa_params.sched_nice;
                
                
 it will be effort to use the retrieved prio for logging instead of INT_MAX or (-1). I don't want to set sched_data->prio to the retrieved value to let the logging show the actual value we used for resetting?

petretudor-arm:
> if (sched_data->prio != THREAD_PRIORITY_UNCHANGED)
>           sa_params.sched_nice = sched_data->prio;
>   else
>           sa_params.sched_nice = _sa_params.sched_nice;

Which is what I suggested in https://github.com/scheduler-tools/rt-app/pull/148#issuecomment-4334680636.

But even then, we only get this behaviour on fair tasks when there is a phases object present. Otherwise it's the current behaviour where you get rt-app's default attributes, since `THREAD_PRIORITY_UNCHANGED` won't make it to the phase's `sched_data`. I think this would be confusing for some users.

deggeman:
True, you're right about the issue of def_prio for def_policy in the phase-less setup. I'm trying:

```switch (tmp_data.policy) {
        case same:
        case other:
        case batch:
        case idle:
               prior_def = THREAD_PRIORITY_UNCHANGED;
                break;
        case fifo:```
       
but I get a segfault right now. Need more time to resolve this ...
```

BTW, how does it work with the system settings on those tasks rt-app creates. Which OS infrastructural bits can set nice and slicelen between task spawning and rt-app doing sched_setattr() on the task?

deggeman:
How do I get meaningful code snippets into those comment fields ???

src/rt-app.c
```diff
@@ -968,22 +969,35 @@ static void __set_thread_sched_other_attrs(thread_data_t *data,
...
+
 	tid = gettid();
-	sa_params.size = sizeof(struct sched_attr);
+	sa_params.size = sizeof(sa_params);
 	sa_params.sched_policy = sched_data->policy;
 	sa_params.sched_priority = __sched_priority(data, sched_data);
+
 	/* In the CFS case, sched_data->prio is the NICE value. */
-	sa_params.sched_nice = sched_data->prio;
+	if (sched_data->prio)
```
deggeman: -       if (sched_data->prio)
+       if (sched_data->prio != THREAD_PRIORITY_UNCHANGED)

src/rt-app.c
```diff
@@ -968,22 +969,35 @@ static void __set_thread_sched_other_attrs(thread_data_t *data,
 		   policy_to_string(sched_data->policy), sched_data->prio,
 		   sched_data->runtime);
 
+	if (sched_data->prio != THREAD_PRIORITY_UNCHANGED || !sched_data->runtime) {
```
petretudor-arm: I think the condition here should be `sched_data->prio == THREAD_PRIORITY_UNCHANGED` otherwise `sched_getattr()` is skipped.

petretudor-arm:
> How do I get meaningful code snippets into those comment fields ???

Markdown syntax for code blocks:

\```
your code here
\```

It also supports language syntax highlighting. I was using

\```C
my code here
\```

deggeman:
Ah, nice! If you press the Code button, it gives me \` \` instead of \``` \```. Thanks!

petretudor-arm:
> BTW, how does it work with the system settings on those tasks rt-app creates. Which OS infrastructural bits can set nice and slicelen between task spawning and rt-app doing sched_setattr() on the task?

From my understanding it is `dup_task_struct()` [[1]], then `cgroup_fork()` [[2]] and finally `sched_fork()` [[3]] in `copy_process()` [[4]], since `pthread_create()` calls `clone3`/`clone`. `sched_fork()` also applies `SCHED_RESET_ON_FORK`, which resets slicelen too.

[1]: https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L2050
[2]: https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L2150
[3]: https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L2193
[4]: https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L1964

deggeman:
> > BTW, how does it work with the system settings on those tasks rt-app creates. Which OS infrastructural bits can set nice and slicelen between task spawning and rt-app doing sched_setattr() on the task?
> 
> From my understanding it is `dup_task_struct()` [[1](https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L2050)], then `cgroup_fork()` [[2](https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L2150)] and finally `sched_fork()` [[3](https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L2193)] in `copy_process()` [[4](https://elixir.bootlin.com/linux/v7.0/source/kernel/fork.c#L1964)], since `pthread_create()` calls `clone3`/`clone`. `sched_fork()` also applies `SCHED_RESET_ON_FORK`, which resets slicelen too.

But we always call pthread_create() and then sched_setattr()? Vincent talked about a "system manager" which could have set a specific runtime for the task, which I assume is a different one than the default se.slice?

E.g. with systemd you can put systemd slice (cgroup placement) and the nice value (scheduler priority) directly in a service unit file:

[Service]
Slice=my-custom.slice
Nice=-5

vingu-linaro:
I faced this case while testing schedqos which sets nice and slice duration.
I'm going to run some tests with this patch

src/rt-app_parse_config.c
```diff
@@ -901,22 +899,13 @@ static sched_data_t *parse_sched_data(struct json_object *obj, int def_policy)
...
+	memcpy(data, &tmp_data,sizeof(sched_data_t));
 
-		log_debug(PIN "key: set scheduler %d with priority %d", new_data->policy, new_data->prio);
-
-		return new_data;
-	}
+	log_debug(PIN "key: set scheduler %d with priority %d", data->policy, data->prio);
 
-	return NULL;
+	return data;
```
vingu-linaro: So you always return a sched_data_t  even if there is no change. We have a mechanism to skip  set_thread_param when there is not change between 2 phases

src/rt-app_parse_config.c
```diff
@@ -901,22 +899,13 @@ static sched_data_t *parse_sched_data(struct json_object *obj, int def_policy)
...
-	if (tmp_data.prio != THREAD_PRIORITY_UNCHANGED ||
-	    tmp_data.runtime || tmp_data.period || tmp_data.deadline ||
-	    tmp_data.util_min != -1 || tmp_data.util_max != -1) {
-		sched_data_t *new_data;
 
-		/* At least 1 parameters has been set in the object */
-		new_data = malloc(sizeof(sched_data_t));
-		memcpy( new_data, &tmp_data,sizeof(sched_data_t));
+	data = malloc(sizeof(sched_data_t));
+	memcpy(data, &tmp_data,sizeof(sched_data_t));
```
petretudor-arm: Should we add a null check here? `memcpy` with a null destination pointer has UB.

deggeman: Yes, you're right. But I would like to avoid this fix in this patch-set since there are other places with the same issue. And this has nothing to do with the actual problem this patch should fix.

src/rt-app_parse_config.c
deggeman: Yeah, this is not necessary. I just have to make sure that I create a sched_data_t for the task in case we have data->sched_data != NULL for the task in case. At least there is a :
```
/* Set scheduling policy and print pretty info on stdout */
log_notice("[%d] Starting with %s policy with priority %d",
                   data->ind, policy_to_string(data->sched_data->policy),
                   data->sched_data->prio);
```
in thread_body().
Let me change this back and fix this differently in the next version.

deggeman:
> I faced this case while testing schedqos which sets nice and slice duration. I'm going to run some tests with this patch

Ah, OK, and schedqos uses netlink to monitor task activities like FORK & EXEC and can then apply QoS params like nice and slicelen.

deggeman:
testcases ran on latest version: 

(~/scripts) e125579 $ ./test_rt-app_sched_setattr.sh 
test_sched_setattr_01 passed
test_sched_setattr_02 passed
test_sched_setattr_03 passed
test_sched_setattr_04 passed
test_sched_setattr_05 passed
test_sched_setattr_06 passed
test_sched_setattr_07 passed
test_sched_setattr_08 passed
test_sched_setattr_09 passed
test_sched_setattr_10 passed

(~/scripts) e125579 $ ./test_rt-app_taskgroups_v1.sh 
test_taskgroup_v1_01 passed
test_taskgroup_v1_02 passed
test_taskgroup_v1_03 passed
test_taskgroup_v1_04 passed
test_taskgroup_v1_05 passed
test_taskgroup_v1_06 passed
test_taskgroup_v1_07 passed
test_taskgroup_v1_11 passed
test_taskgroup_v1_12 passed
test_taskgroup_v1_13 passed
test_taskgroup_v1_14 passed
test_taskgroup_v1_15 passed
test_taskgroup_v1_16 passed
test_taskgroup_v1_21 passed
test_taskgroup_v1_22 passed
test_taskgroup_v1_23 passed
test_taskgroup_v1_24 passed
test_taskgroup_v1_25 passed
test_taskgroup_v1_26 passed
test_taskgroup_v1_31 passed
test_taskgroup_v1_32 passed
test_taskgroup_v1_33 passed
test_taskgroup_v1_34 passed
test_taskgroup_v1_35 passed
test_taskgroup_v1_36 passed
test_taskgroup_v1_37 passed
test_taskgroup_v1_38 passed
test_taskgroup_v1_41 passed
test_taskgroup_v1_42 passed
test_taskgroup_v1_43 passed
test_taskgroup_v1_44 passed
test_taskgroup_v1_51 passed
test_taskgroup_v1_52 passed
test_taskgroup_v1_53 passed
test_taskgroup_v1_54 passed
test_taskgroup_v1_61 passed
test_taskgroup_v1_62 passed
test_taskgroup_v1_63 passed
test_taskgroup_v1_64 passed
test_taskgroup_v1_65 passed
test_taskgroup_v1_66 passed
test_taskgroup_v1_67 passed
test_taskgroup_v1_68 passed
test_taskgroup_v1_71 passed
test_taskgroup_v1_72 passed
test_taskgroup_v1_73 passed
test_taskgroup_v1_74 passed
test_taskgroup_v1_75 passed
test_taskgroup_v1_76 passed

[indiv_setting_nice_runtime_for_normal_batch_test.tgz](https://github.com/user-attachments/files/27521970/indiv_setting_nice_runtime_for_normal_batch_test.tgz)

vingu-linaro:
@deggeman Thanks for the update, I will review your latest version beg of next week

vingu-linaro:
I have the log below when I don't set a prio. The end result is ok because there is no scheduler change but the log is weird with 2147483647 for THREAD_PRIORITY_UNCHANGED :

> [rt-app] <notice> [0] Starting with SCHED_OTHER policy with priority 2147483647
> [rt-app] <debug> [0] setting scheduler SCHED_OTHER priority 2147483647

deggeman:
```C
(1) [rt-app] [0] Starting with SCHED_OTHER policy with priority 2147483647
(2) [rt-app] [0] setting scheduler SCHED_OTHER priority 2147483647

thread_body(void *arg)

  log_notice(..., data->sched_data->prio)                               <-- (1)

  set_tread_param(data, data->sched_data->prio

    _set_thread_cfs()

     __set_thread_policy_priority(data, sched_data)

     __set_thread_sched_other_attrs(data, sched_data)

      __log_policy_priority_change()

        log_debug("[%d] setting scheduler %s priority %d", data->ind,   <-- (2)
                  policy_to_string(sched_data->policy),
                  sched_data->prio);


I would have to base the logs onto the actual 'struct sched_param' and 'struct sched_attr' information. 

We can remove (1) because there are already 2 notices indicating that a thread has been started and named:

[rt-app] <notice> thread_data_set_unique_name 0 t0-0
[rt-app] <notice> [0] starting thread ...

and then put (2) __set_thread_policy_priority(), where policy and priority is actually set.

For setting SCHED_OTHER/SCHED_BATCH nice and/or runtime value, we already have a log_debug()/log_ftrace() which I would simply move after the actual sched_setattr().

[rt-app] <debug> [0] setting scheduler SCHED_BATCH priority 0
[rt-app] <debug> [0] setting scheduler SCHED_BATCH nice=0 runtime=40000000

Let me create a patch for this.
```

vingu-linaro:
Looks good for me.
Thanks

PR #150, opened 2026-05-05
rt-app: remove duplicate t_start init

vingu-linaro:
t_start is init twice back to back. The 2nd one refers to the timers which use t_zero that is initialized by 1st thread.


jlelli:
Ack! Thanks!

